PR #196, opened 2018-07-05
Support the running of CombinedFSkObject in the runner node

ahmadswaid:


miguelalba:
Please remove println in RunnerNodeModel.

PR #199, opened 2018-07-17
Sbml writing

ahmadswaid:


ahmadswaid:
I'm sure, I have done the code formatting before committing.
It seems to be an issue in your eclipse

PR #415, opened 2020-05-04
Deploy check

schuelet:
All the necessary code is in the .yml file. the bash script I added last week can be deleted.


.travis.yml
```diff
@@ -4,7 +4,143 @@ jdk:
...
+-
+- echo "INSTALL OLD FSK-LAB TEMPORARILY"
+- knime_4.1.2/knime -application org.eclipse.equinox.p2.director -repository "$KNIME41,$OLD_FSK" -installIU de.bund.bfr.knime.fsklab.feature.feature.group
+- knime_4.1.2/knime -application org.eclipse.equinox.p2.director -lir
+- echo "REMOVING OLD FSK-LAB"
+- knime_4.1.2/knime -application org.eclipse.equinox.p2.director -uninstallIU de.bund.bfr.knime.fsklab.feature.feature.group
+- knime_4.1.2/knime -application org.eclipse.equinox.p2.director -lir
+- echo "REINSTALLING NEW FSK-LAB"
+- knime_4.1.2/knime -application org.eclipse.equinox.p2.director -repository "$KNIME41,$NEW_FSK" -installIU de.bund.bfr.knime.fsklab.feature.feature.group
+- knime_4.1.2/knime -application org.eclipse.equinox.p2.director -lir
```
miguelalba, line 123: These commands listing the installed plugins are not needed and produce noise here. They are only used in manual tests. They should not be used in a "batch" mode.

.travis.yml
```diff
@@ -4,7 +4,143 @@ jdk:
   - openjdk8
 script:
 - xvfb-run mvn verify -B -Pjacoco
-- bash deployment_check.sh
```
miguelalba, line 7: I do like having a separate script for the test. It keeps the yaml file clean. Please do not run this for every build like you are doing here. Use instead a conditional build for your branch. https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#conditional-builds. You would need to maintain a branch mirroring the master that would execute this check and could be triggered by a cron job.

Another approach and with no maintenance could be to detect if the build was activated by a cron job. This approach would use any branch actually and does not need to keep updated a separate branch. See https://docs.travis-ci.com/user/cron-jobs/#detecting-builds-triggered-by-cron

PR #430, opened 2020-06-17
Joiner dirty fix

schuelet:
- support for multiple level joining
- support for additional simulations for combined models
- default simulation based on sub-models
- new system to track parameters through the joining tree
- support for join command 
- fixed bug of joiner view, in which switching between join connections removed the join command
- open issues:
  - reader & writer not tested yet for combined models
  - join command in outport only visible for top level join
  - joiner view not loaded correctly into view if node was already executed 
  - resetting joiner node and creating a joined model doesn't result in a correct combined object


de.bund.bfr.knime.fsklab.nodes.tests/src/de/bund/bfr/knime/fsklab/nodes/JoinerNodeUtilTest.java
```diff
@@ -3,81 +3,239 @@
...
 
 @SuppressWarnings("static-method")
 public class JoinerNodeUtilTest {
 
+	
 	@Test
-	public void testResolveParameterNamesConflict() {
+	public void testGetOriginalParameterNames_CorrectOutputParameters_shouldReturnPresent() {
+		
+		LinkedHashMap<String, String> originalOutputParams = new LinkedHashMap<String, String>();
```
miguelalba, line 19: format

miguelalba, line 19: Types not needed because of type inference in diamond operator since Java 7

Issue #119, opened 2018-03-06
Integrate FSK Simulation Configurator features into FSK Runner node


mfilter:

miguelalba:
Simulator node:
![simulator](https://user-images.githubusercontent.com/19906968/46288532-5066bd00-c586-11e8-8070-0f5f43543ff1.png)

Runner node:
![runner](https://user-images.githubusercontent.com/19906968/46288533-50ff5380-c586-11e8-8699-d9ecdc2b32c0.png)

Some considerations:
* The simulator node is JavaScript-based while the runner is a pure Java node. To integrate the simulator into the runner and be able to use on the server it will be necessary to rewrite the runner as a JavaScript node.
* If we were to remove the simulator then the user would always need to run the model in order to create a simulation. The user could not just create the simulation like now.

PR #527, opened 2020-09-01
#119: Moved METADATA_COMMAND_VALUE to NodeUtils

schuelet:


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/ReaderNodeModel.java
```diff
@@ -77,6 +77,8 @@
 import de.bund.bfr.knime.fsklab.FskPortObjectSpec;
 import de.bund.bfr.knime.fsklab.FskSimulation;
 import de.bund.bfr.knime.fsklab.JoinRelation;
+import de.bund.bfr.knime.fsklab.nodes.JoinerNodeUtil;
```
miguelalba, line 80: We should remove the import to JoinerNodeUtil as it is not used anymore.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/ReaderNodeModel.java
```diff
@@ -77,6 +77,8 @@
...
+import de.bund.bfr.knime.fsklab.nodes.NodeUtils;
```
miguelalba, line 81: We should remove the import to NodeUtils since it is not needed because ReaderNodeModel is in the same package.

miguelalba (changes requested):
Very good! Please apply the requested changes and the rest is fine.

miguelalba (approved):

PR #531, opened 2020-09-01
Task/513 Move files of the CreatorNode

miguelalba:
Move the files to a package de.bund.bfr.knime.fsklab.v1_9.creator to not conflict with the ICPMF deprecated node.


miguelalba:
The branch used to open has an old commit (Make methods in ScriptHandler public) that is already included in story/512. Another PR should replace this with only the commit to move the creator node files.

Issue #568, opened 2020-09-03
Move 1.7.2 creator node to deprecated project


miguelalba:

schuelet:
Tested on Linux

PR #581, opened 2020-09-03
moved creator to src-1_7_2

schuelet:
#568 had to also move some Runner classes because the ScriptHandler
needed it


miguelalba (approved):
Sehr gut

PR #605, opened 2020-09-09
Bugfix/597

schuelet:


schuelet:
delete branch, replaced by task/597

PR #627, opened 2020-09-11
Clean deprecated

miguelalba:
This branch fixes the 1.7.2 JS components of the editor, simulator and joiner nodes. The two first are completely refactored but the joiner dependencies are pretty weak and coupled. Almost every attempt to replace them with the dependencies from KNIME or de.bund.bfr.knime.js produces errors and crashes the joiner.


mfilter:
Couldn't Ahmad do the refactoring of the Joiner? He has to work on it anyway. As long as the deprecated Joiner from the ICPMF release is still working we can live for the moment with a non-functional joiner.

schuelet (approved):

PR #674, opened 2020-09-24
Fix packages installation on Mac

miguelalba:
The packages are installed on Mac like on Linux, by skipping minCRAN and using install.packages to the .fsk/library folder.


schuelet (approved):

mfilter:
I think this should undergo a review / QA process before we close this ticket

miguelalba:
Unfortunately this is not an actual issue but a pull request but it shows as an issue. I cannot re open it or move it to another lane. I could instead link it to a related ticket like https://github.com/RakipInitiative/ModelRepository/issues/138 and leave that at QA.

mfilter:
OK for me

PR #680, opened 2020-09-28
convert input object into JSONDataTable.

ahmadswaid:


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/fskdbview/FSKDBViewNodeModel.java
```diff
@@ -185,9 +185,41 @@ public void saveCurrentValue(NodeSettingsWO content) {
...
+   * A helper method for creating an empty table in the case of empty input port.
+   * 
+   * @param ExecutionContext exec.
+   * @return an empty BufferedDataTable instance.
+   */
+  private BufferedDataTable createEmptyTable(final ExecutionContext exec) {
+    BufferedDataContainer container =
+        exec.createDataContainer(new DataTableSpecCreator().createSpec());
+    container.close();
+    BufferedDataTable out = container.getTable();
```
miguelalba, line 221: We should not declare local variables and return them immediately after. We should return instead *container.getTable()*. https://rules.sonarsource.com/java/RSPEC-1488?search=local%20variables%20should%20not%20be%20declared%20and%20then%20immediately%20returned

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/fskdbview/FSKDBViewNodeModel.java
```diff
@@ -185,9 +185,41 @@ public void saveCurrentValue(NodeSettingsWO content) {
   @Override
   protected PortObject[] performExecute(PortObject[] inObjects, ExecutionContext exec)
       throws Exception {
+    PortObject outport = inObjects[0];
```
miguelalba, line 188: This is more of a personal preference, but in the former editor and joiner the input port was also reused as output port and it got complicated when the node was finished and the execute method grew big. I would suggest here to explicitely define an input and output ports so that later when this method grows it will be still be clear.

miguelalba (changes requested):
Pretty good.

miguelalba (approved):

PR #693, opened 2020-10-05
restored line of code. Without it, the node crashed.

schuelet:


miguelalba:
@schuelet I think we can close this now as the the branch was already merged.

PR #702, opened 2020-10-18
648#four optional input port for the joiner runnable writable

ahmadswaid:


miguelalba (approved):
It is fine. There are small things I would like changed but I do not want to slow down the work with the node. The models in the test pipeline are running as usual.

PR #703, opened 2020-10-25
Joiner reorder use current value as default mouseover tooltip

ahmadswaid:


miguelalba:
Hi Ahmad, this PR changes over 1600 lines of code and I find really hard to check and test everything and not missing bugs. Do you think we could split it into smaller PRs? Some ideas to split it are:
- isEmpty method in JoinerViewValue
- changes to the representation and value classes
- changes to settings. I think this changes might not be split from the changes to the representation and value classes
- changes to JS and CSS (over 1000 lines changed).

Regarding the settings in the representation, value and config for each of the inputs models, from the first to the fourth model, it is always the same. You work with the model names and math for each of them and most of the code is identical and written four times. We could use a POJO inside the representation and value that wraps this data like FskSimulation in the simulator node and try to avoid the repeated code.

ahmadswaid:
Unfortunately, I can't divide it more than two (Java Part and the JS one). I don't know ho to test if you don't have the all changes.
the changed Java classes are interdependent and changes has to be merged together, otherwise the build will fail.
BTY, I like the idea of POJO here, implemented it in the next PR.

PR #714, opened 2020-11-04
Databases should be password-protected (java:S2115)

miguelalba:
Add user and password to the RAKIP vocabularies database.

## java:S2115
Databases should always be password protected. The use of a database
connection with an empty password is a clear indication of a database
that is not protected.
This rule flags database connections with empty passwords.


schuelet (approved):

miguelalba:
There is a conflict but FskService is also changed in https://github.com/SiLeBAT/FSK-Lab/pull/715. I will wait for that PR to be merge before fixing the conflict here.

schuelet (approved):

PR #716, opened 2020-11-08
migrate joiner which has no suffix feature in parameters and connections

ahmadswaid:


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -406,20 +404,31 @@ protected void loadJsonSetting() throws IOException, CanceledExecutionException
...
       representation.setFirstModelParameters(nodeSettings.firstModelParameters);
     }
     if (nodeSettings.secondModelParameters != null) {
       representation.setSecondModelParameters(nodeSettings.secondModelParameters);
     }
+
+
+    if (!JoinerNodeUtil.parametersNeedUpdate(newFirstModelParameters)) {
+      representation.updateParameters(newFirstModelParameters, newSecondModelParameters);
+      viewValue.updateParameters(newFirstModelParameters, newSecondModelParameters);
```
miguelalba, line 425: Do not we need to set the viewValue.modelMetadata when the parameters need to be updated? What happens to the metadata in this case?

miguelalba (changes requested):

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeUtil.java
```diff
@@ -109,8 +110,38 @@ public static void addIdentifierToParameters(
...
+        
+        for (Parameter param : params) {
+          String paramID = param.getId();
+          char currectLastChar = paramID.charAt(paramID.length()-1);
+          if(currectLastChar == lastChar) {
+            if (!(paramID.endsWith(JoinerNodeModel.SUFFIX_FIRST)
+                || (paramID.endsWith(JoinerNodeModel.SUFFIX_SECOND)))) {
+              needsUpdate = true;
+            }
+          }else {
```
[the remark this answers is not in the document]
ahmadswaid: - I think the assumption is wrong, the first and second iteration will not change the `needsUpdate` to true because they don't end with `JoinerNodeModel.SUFFIX_FIRST, JoinerNodeModel.SUFFIX_SECOND`.
- The change to false before break is done to cover a cases like: `log10_D1, log10_D1, dog` and this is covered by unittest.
- by reviewing the code I improve it to return the method on finding the first parameter doesn't end with either suffixes.
- it's a good catch for that the iteration has to start from the second item. solved in commit

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
ahmadswaid: in the case that the parameters need to be updated the metadata will be generated again and will not be taken from saved setting.

miguelalba (approved):

PR #719, opened 2020-11-12
Support early versions of setting files

ahmadswaid:


de.bund.bfr.knime.fsklab.deprecatednodes/src-1_7_2/de/bund/bfr/knime/fsklab/nodes/v1_7_2/joiner/JoinerNodeModel.java
```diff
@@ -336,14 +351,26 @@ private void creatRelationList(String relation, JoinerViewValue joinerProxyValue
           jR.setLanguage_written_in(sourceTargetRelation.getString("language_written_in"));
         }
         if (sourceTargetRelation.containsKey("sourceParam")) {
-          jR.setSourceParam(getObjectFromJson(sourceTargetRelation.get("sourceParam").toString(),
```
miguelalba, line 339: I am lost here. So the whole *creatRelationList* loads a JsonString as a Javax JsonObject, processes directly simple properties (command and language written in) and for the source and target parameters they are converted to strings that are later deserialised with Jackson. Also, in the case that the content in the sourceParam or targetParam nodes is invalid this is using the whole raw content as the id of a new empty parameter (catch clause). Is this correct?

I would simplify this by data type binding with Jackson or simply reading the JSON string as a JsonTree and iterate like you do here. The nodes for the source and target parameters could be handled individually and deserialised to the Parameter classes. Do we have examples or tests with these join relations?

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -344,11 +354,27 @@ protected void loadJsonSetting() throws IOException, CanceledExecutionException
 
     if (flowVariables.containsKey("JoinRelations.json")) {
       String connectionString = flowVariables.get("JoinRelations.json").getStringValue();
-      nodeSettings.connections = MAPPER.readValue(connectionString, JoinRelation[].class);
+      try {
+        nodeSettings.connections = MAPPER.readValue(connectionString, JoinRelation[].class);
+      }catch(Exception ex) {
+        OldJoinerRelation[] oldRelations= MAPPER.readValue(connectionString, OldJoinerRelation[].class);
+        Function<OldJoinerRelation, JoinRelation> funcMigrateJoinerRelation= (OldJoinerRelation e)-> {return e.getNewJoinRelation();};
+        nodeSettings.connections  = Arrays.asList(oldRelations).stream().map(funcMigrateJoinerRelation
```
miguelalba, line 362: We could stream the array directly with Arrays.stream instead of converting the array to a list and then stream it. See https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#stream-T:A-

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -344,11 +354,27 @@ protected void loadJsonSetting() throws IOException, CanceledExecutionException
...
       File configFile = new File(settingFolder, "JoinRelations.json");
       if (configFile.exists()) {
-        nodeSettings.connections = MAPPER.readValue(configFile, JoinRelation[].class);
+        try {
+          nodeSettings.connections = MAPPER.readValue(configFile, JoinRelation[].class);
+        } catch (Exception ex) {
+          OldJoinerRelation[] oldRelations =
+              MAPPER.readValue(configFile, OldJoinerRelation[].class);
+          nodeSettings.connections =
+              Arrays.asList(oldRelations).stream().map((OldJoinerRelation e) -> {
```
miguelalba, line 374: Same as above.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -360,7 +386,7 @@ protected void loadJsonSetting() throws IOException, CanceledExecutionException
         nodeSettings.modelMetaData = FileUtils.readFileToString(configFile, StandardCharsets.UTF_8);
       }
     }
-
+    List <Parameter> firstParams = new ArrayList<Parameter>();
```
miguelalba, line 389: Unnecessary Parameter type with diamond operator.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -418,19 +452,37 @@ protected void loadJsonSetting() throws IOException, CanceledExecutionException
...
+   * de.bund.bfr.metadata.swagger.parameter
+   */
+  private static List<Parameter> getModelParametersfromOldFile(File configFile)
+      throws FileNotFoundException, IOException, ParseException {
+    JSONParser parser = new JSONParser();
+    Object obj = parser.parse(new FileReader(configFile));
+    JSONObject jsonObject = (JSONObject) obj;
+    JSONArray parameterList = (JSONArray) jsonObject.get("parameter");
+    return MAPPER.readValue(parameterList.toJSONString(), new TypeReference<List<Parameter>>() {});
+  }
```
miguelalba, line 484: The biggest problem in this method is the leak of the FileReader that is not closed (480).

This method is similar to creatRelationList of the 1.7.2 JoinerNodeModel. We are using 2 JSON libraries. First we parsed the passed configuration file with JSON simple, get the parameters, convert them to a string which is finally deserialised to a Parameter objects with Jackson.

1. How does this JSON configuration file look like? Can we deserialise with data binding to a POJO? If this does not exist we can just create it and drop this whole method.
2. In case 1 takes a lot of effort we could do everything in this method with Jackson. The contents of the file can be read as a JsonTree and its parameters can be deserialised manually. [treeToValue](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#treeToValue(com.fasterxml.jackson.core.TreeNode,%20java.lang.Class))

We should not forget to close the FileReader. With Jackson we would not need it anyway https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#readTree(java.io.File).

miguelalba (changes requested):

de.bund.bfr.knime.fsklab.deprecatednodes/src-1_7_2/de/bund/bfr/knime/fsklab/nodes/v1_7_2/joiner/JoinerNodeModel.java
ahmadswaid: I understand your concerns but There are two types of setting for the JoinRelations.json: 
one with the parameter ID only and the other (I think it is the older version) is hosting the whole parameter information.
I used Jackson to do the conversion into a Parameter Object and if it fails the catch will try to create a parameter with the existing parameter ID in the file.
please see: https://app.zenhub.com/files/241333403/fb898916-df34-48cd-ac4b-d5810b7be5f9/download
and https://seafile.bfr.berlin/f/6e84bfd933074036bd3e/

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -476,11 +477,8 @@ protected void loadJsonSetting() throws IOException, CanceledExecutionException,
    */
   private static List<Parameter> getModelParametersfromOldFile(File configFile)
       throws FileNotFoundException, IOException, ParseException {
-    JSONParser parser = new JSONParser();
-    Object obj = parser.parse(new FileReader(configFile));
-    JSONObject jsonObject = (JSONObject) obj;
-    JSONArray parameterList = (JSONArray) jsonObject.get("parameter");
-    return MAPPER.readValue(parameterList.toJSONString(), new TypeReference<List<Parameter>>() {});
+    JsonNode nodes = MAPPER.readTree(configFile);
+    return new ArrayList<>(Arrays.asList(MAPPER.treeToValue(nodes.get("parameter") , Parameter[].class)));
```
miguelalba, line 480: It is not necessary to copy the list. We can return the list from Arrays.asList `Arrays.asList(MAPPER.treeToValue(nodes.get("parameter") , Parameter[].class))`

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -68,6 +68,7 @@
 import org.knime.js.core.node.AbstractSVGWizardNodeModel;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
```
miguelalba, line 71: Please check unused imports or just automatise it with Eclipse: Cmd+Shift+O or Ctrl+Shift+O.

miguelalba (changes requested):

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
ahmadswaid: the list should be mutable because I remove some elements from it in other places.
some files contain extra element which has to be removed before usage.

miguelalba (approved):

PR #722, opened 2020-11-17
Running Joiner Component

ahmadswaid:


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -596,8 +599,8 @@ private CombinedFskPortObject createCombinedFskPortObject(FskPortObject jFirstIn
   }
 
   private static FskPortObject createEmptyFSKObject() throws IOException {
-
-    FskPortObject fskPortObject = new FskPortObject(Optional.empty(), "", Collections.emptyList());
+    
+    FskPortObject fskPortObject = new FskPortObject(Optional.of(new FilesEnvironmentManager(new String[0])), "", Collections.emptyList());
```
miguelalba, line 603: This challenges the approach of the environment managers. An empty manager object is more transparent as it explicitly describes a lack of environment. With an empty files environment or environment with 0 files we are hiding the fact that this environment is empty and enabling some other client code to use a manager, probably because it has not been updated to support the lack of environment.
I would rather help to update the client code to support the case of models with no environment rather than including this workaround.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -631,10 +634,10 @@ private static FskPortObject createEmptyFSKObject() throws IOException {
       JoinerModelsData joinerModelsData = value.joinerModelsData;
       if (joinerModelsData.firstModel != null && joinerModelsData.firstModel.length > 0
           && StringUtils.isNotEmpty(joinerModelsData.firstModel[0])) {
-        outObj.modelMetadata = MAPPER.readValue(value.modelMetaData,
-            SwaggerUtil.modelClasses.get(firstInputPort.modelMetadata.getModelType()));
+        
+        
         FskPortObject jFirstInputPort = getFSKObjectFromStringArray(
-            new FskPortObject(Optional.empty(), "", new ArrayList<String>()),
+            new FskPortObject(firstInputPort.getEnvironmentManager(), "", new ArrayList<String>()),
```
miguelalba, line 640: We could avoid the creation of the empty lists with Collections.emptyList. This list is immutable but it is fine and desirable as we do not modify the list of packages. When a model is edited a new port object is returned at the output (not the input one edited). See https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyList--

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -832,8 +843,10 @@ private static FskPortObject getFSKObjectFromStringArray(FskPortObject portObjec
       String modelType) throws JsonMappingException, JsonProcessingException, IOException {
 
     portObject.modelMetadata = MAPPER.readValue(model[0], SwaggerUtil.modelClasses.get(modelType));
-    portObject.setModel(MAPPER.readValue(model[1], String.class));
-    portObject.setViz(MAPPER.readValue(model[2], String.class));
+    if(!StringUtils.isEmpty(model[1]))
```
miguelalba, line 846: Since we are using StringUtils we could use isNotEmpty instead of negating the conditions.

miguelalba (changes requested):

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
ahmadswaid: actually, I do change the list of packages in line 855 so I need it to be mutable list.

miguelalba: Yes and this shows in my opinion another problem (copied method)

```java
  private static FskPortObject getFSKObjectFromStringArray(FskPortObject portObject, String[] model,
      String modelType) throws JsonMappingException, JsonProcessingException, IOException {

    portObject.modelMetadata = MAPPER.readValue(model[0], SwaggerUtil.modelClasses.get(modelType));
    if(StringUtils.isNotEmpty(model[1]))
      portObject.setModel(MAPPER.readValue(model[1], String.class));
    if(StringUtils.isNotEmpty(model[2]))
      portObject.setViz(MAPPER.readValue(model[2], String.class));
    portObject.simulations.clear();
    portObject.packages.clear();
    portObject.simulations
        .addAll(MAPPER.readValue(model[3], new TypeReference<List<FskSimulation>>() {}));

    portObject.packages.addAll(MAPPER.readValue(model[4], new TypeReference<List<String>>() {}));

    return portObject;
  }
```

In this method we are taking a passed port object and modifying it. It would be better to leave as much as possible of the port object immutable and return instead a new port object in this method. The first use of this is in the JoinerNodeModel

```java
        FskPortObject jFirstInputPort = getFSKObjectFromStringArray(
            new FskPortObject(firstInputPort.getEnvironmentManager(), "", new ArrayList<String>()),
            joinerModelsData.firstModel, joinerModelsData.firstModelType);
```
Here we create a port object with few data and then pass it to getFSKObjectFromStringArray which fills more data in the port object from the array. Would not be simpler to initialise the port object in a single step. I will just open a ticket for this change and for now we can use the new empty list.

ahmadswaid: method improved

miguelalba (approved):

PR #730, opened 2020-11-23
Joining algorithm fixes

miguelalba:
Joining algorithm returns a GenericModel instead of JsonNode and several issues related to null and empty properties are fixed.

Example:
```java
  public static void main(String[] args) throws IOException {
	    
	    ObjectMapper mapper = new ObjectMapper().registerModule(new ThreeTenModule());
	    
	    JsonNode exampleModel = mapper.readTree(new File("files/marcel/nautaCPM.json"));
	    JsonNode exampleModel2 = mapper.readTree(new File("files/marcel/medChallDRM.json"));
	    JsonNode[] models = { exampleModel, exampleModel2 };
	    
	    URL url = new URL("http://localhost:9999/joinMetadata");
	    HttpURLConnection conn = (HttpURLConnection) url.openConnection();
	    conn.setRequestMethod("POST");
	    conn.setRequestProperty("Content-Type", "application/json; utf-8");
	    conn.setRequestProperty("Accept", "application/json");
	    conn.setDoOutput(true);
	    
	    try (OutputStream outputStream = conn.getOutputStream()) {
	      mapper.writeValue(outputStream, models);
	    }
	    
	    // Response
	    int status = conn.getResponseCode();
	    
	    try (InputStream stream = conn.getInputStream()) {
//	    	JsonNode output = mapper.readTree(stream);
	    	GenericModel output = mapper.readValue(stream, GenericModel.class);
	    	System.out.println(output);
	    }
	    
	    conn.disconnect();
	  }
```

### DIFF
--- a/de.bund.bfr.knime.fsklab.service/src/de/bund/bfr/knime/fsklab/service/FskService.java
+++ b/de.bund.bfr.knime.fsklab.service/src/de/bund/bfr/knime/fsklab/service/FskService.java
@@ -28,10 +28,13 @@ import org.h2.tools.DeleteDbFiles;
 import org.knime.core.node.NodeLogger;
 import org.osgi.framework.Bundle;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.threetenbp.ThreeTenModule;
 import com.google.gson.Gson;
 
+import de.bund.bfr.metadata.swagger.GenericModel;
 import de.bund.bfr.rakip.vocabularies.data.AccreditationProcedureRepository;
 import de.bund.bfr.rakip.vocabularies.data.AvailabilityRepository;
 import de.bund.bfr.rakip.vocabularies.data.BasicProcessRepository;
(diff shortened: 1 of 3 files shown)


ahmadswaid (approved):
it's fine.

ahmadswaid:
[exception.txt](https://github.com/SiLeBAT/FSK-Lab/files/5592589/exception.txt)

miguelalba:
The exception comes from two creation dates joined

> Cannot deserialize value of type `org.threeten.bp.LocalDate` from String \"2018-07-16_2020-08-04\": Text '2018-07-16_2020-08-04' could not be parsed

The creation date is a single property and unlike the modification dates that are contained in an array, I cannot join them. With the arrays I am just merging them into a single array, but for date strings I cannot join them like with the rest of strings because I break the date format (error). So I need to set only of the creation dates of the two models joined. Would that be enough for you?

ahmadswaid:
> The exception comes from two creation dates joined
> 
> > Cannot deserialize value of type `org.threeten.bp.LocalDate` from String "2018-07-16_2020-08-04": Text '2018-07-16_2020-08-04' could not be parsed
> 
> The creation date is a single property and unlike the modification dates that are contained in an array, I cannot join them. With the arrays I am just merging them into a single array, but for date strings I cannot join them like with the rest of strings because I break the date format (error). So I need to set only of the creation dates of the two models joined. Would that be enough for you?

I think, it's a thing the "metadata joining algorithm" has to decide and not the joiner.
As an Idea, We can consider the current date as the creation date

miguelalba:
I added a fix to only take one creation date from the models. First it takes the creation date of the first model and if missing, it takes the one from the second model. I cannot join the creation dates because I would break the format: <some date>_<another_date> is a not a valid date format and this why you got the exception.

de.bund.bfr.knime.fsklab.service/src/de/bund/bfr/knime/fsklab/service/FskService.java
```diff
@@ -78,275 +81,277 @@
...
+
+			try {
+				JsonNode inputMetadata = jsonTransformer.MAPPER.readTree(req.body());
+				String targetClass = req.params("targetModelClass");
+				JsonNode convertedMetadata = utils.convertModel(inputMetadata, targetClass);
+
+				res.status(200);
+				return convertedMetadata;
+			} catch (Exception err) {
+				res.status(400);
```
ahmadswaid, line 165: This could be improved in the future since 400 is not describing always  good what really happened.
I would prefer to use 500 error code rather than using 400.
the 400 can be used to notify that the endpoint received not well formatted request parameters.
But for now It's fine and the algorithm works for my real testing.

ahmadswaid (approved):

de.bund.bfr.knime.fsklab.service/src/de/bund/bfr/knime/fsklab/service/FskService.java
miguelalba: Thanks! I will open an issue for this.

PR #735, opened 2020-11-29
Merging simulation strategies

ahmadswaid:


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -716,7 +715,11 @@ private static FskPortObject createEmptyFSKObject() throws IOException {
 
 
       // Create default simulation out of parameters metadata
-      JoinerNodeUtil.createDefaultSimulation(outObj);
+      if(!joinerModelsData.joinedSimulation.isEmpty()) {
+        outObj.simulations.addAll(joinerModelsData.joinedSimulation);
+      }else {
+        JoinerNodeUtil.createDefaultSimulation(outObj);
```
miguelalba, line 721: This is fine and I can approve it but I do not like that in case that there is a simulation we simply set it here, but when there are no simulations we create it from the metadata and set it the createDefaultSimulation. Also createDefaultSimulation does not only create the simulation but also sets the selected simulation index. I would prefer that the method only created the simulation and we would set them here for both cases. This is something I think fits better in a future ticket.

miguelalba (approved):

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
ahmadswaid: the creation of a default simulation happens in the JavaScript Side as a merge of the model's default simulations and that one in the selected one by default.

PR #740, opened 2020-12-09
Fix table deselection use local fsk working directory

ahmadswaid:


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v1_9/joiner/JoinerNodeModel.java
```diff
@@ -855,13 +856,31 @@ private static String getObjectAsJSONString(Object object) {
...
     try(FileOutputStream fileOutputStream = new FileOutputStream(fileName)){
       fileOutputStream.getChannel().transferFrom(readableByteChannel, 0, Long.MAX_VALUE);
     }
   }
+  
+  /**
+   * 
+   * @return the path to .fsk folder in the hosting workflow and create it if not exist.
+   */
+  private static String getFSKDirectoryPath() {
```
miguelalba, line 876: I do not like to create folders inside the workflow. This is using low-level implementation of KNIME. We had problems already of settings stored in workflow that were broken as soon as the nodes were wrapped into metanodes/components as the paths in the workflow folder change. If you insist in creating a folder inside the workfow folder you could look into getWorkflowTempDir() of org.knime.core.util.FileUtil. By the way, the folder created here is not deleted yet.

Still, I would prefer to not create folders yet and use instead the environment manager. For example, if most of the models do not have resources (only scripts) then the working directory would not be necessary and created with FilesEnvironmentManager. With this approach instead empty folders are created for every model.

miguelalba (changes requested):

miguelalba (approved):

PR #750, opened 2021-02-01
Frontend maven plugin support

ahmadswaid:
- Move https://github.com/sod-dev/bfr-landingpage to FSKLab (No extra files or changes are added to these file after moving).
- Add support of automatic build (grunt) of the shared JS lib using frontend-maven-plugin.
- this can be tested by applying `mvn clean verify` command on the level of de.bund.bfr.knime.js module


miguelalba:
We have a conflict with the app.js. I think we should keep it away from the repository and generate it with the build.

ahmadswaid:
> We have a conflict with the app.js. I think we should keep it away from the repository and generate it with the build.

File is removed from the branch and added that to a .ignore file

PR #764, opened 2021-02-18
support cascade joining

ahmadswaid:
This is necessary when doing a join of first two models then with using another joiner to combine with a third one. 
in this case the runner was failing.


schuelet:
Hi Ahmad, 
so this does work fine, but have you ever considered that the second model to be joined is a combined model? The Runner will fail in that case, but that might be my problem not yours.
[joiningCombinedModels.zip](https://github.com/SiLeBAT/FSK-Lab/files/6001436/joiningCombinedModels.zip)

schuelet (approved):

ahmadswaid:
great, I will check that.

PR #768, opened 2021-02-19
fixed a bug, where simulation files are read into workingDirectory

schuelet:
In the Reader, simulation (e.g. simulations/defaultsimulation.r) files would be considered part of the
"entries" to be loaded into the workingdirectory (environment). This bug
has the consequence, that the simulation.r (or .py) files end up in the
root folder of the written fskx file.


schuelet:
@miguelalba This would of course mean, that any files in subfolders are from now on ignored. Did we ever say we supported files in subfolders?

miguelalba (approved):

PR #773, opened 2021-02-22
fixed some issues if json data is large array

schuelet:
Handler now creates a temp file that stores JSON data. I experienced issues with R not being able to read large json strings, so now it should be much more stable.

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/RJsonHandler.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/RJsonHandler.java
@@ -90,12 +90,12 @@ public class RJsonHandler extends JsonHandler {
       if (sourceParam.equals(param.getMetadata().getId())) {
         // store data in temp file because moving big arrays between controller and Java
         // doesn't seem to work properly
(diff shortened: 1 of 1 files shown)


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/RJsonHandler.java
```diff
@@ -87,15 +88,22 @@ public void loadParametersIntoWorkspace(String parameterJson, String sourceParam
     String language = parameterData.getGeneratorLanguage();
     for (DataArray param : parameterData.getParameters()) {
       if (sourceParam.equals(param.getMetadata().getId())) {
+        // store data in temp file because moving big arrays between controller and Java
+        // doesn't seem to work properly
+        File temp_data = FileUtil.createTempFile("data", "json");
```
miguelalba, line 93: Local variable and method parameter names should comply with a naming convention (java:S117).

Shared naming conventions allow teams to collaborate effectively. This rule raises an issue when a local variable or function parameter name does not match the provided regular expression.

Noncompliant Code Example

With the default regular expression ^[a-z][a-zA-Z0-9]*$:

miguelalba (changes requested):

PR #774, opened 2021-02-22
removed the 100 cap on simulation parameter sliders

schuelet:


de.bund.bfr.knime.js/src/js/lib/bootstrap-touchspin.js
```diff
@@ -34,7 +34,7 @@
 
     var defaults = {
       min: 0, // If null, there is no minimum enforced
```
ahmadswaid, line 36: @schuelet  do you see that we have to open the limitation on the min value?

ahmadswaid (approved):

PR #775, opened 2021-02-22
joiner area zoom in/out

ahmadswaid:
zoom in/out buttons are added / mouse wheel zoom support to the joiner area


schuelet:
The zooming works very well. 
But in the Component, the Joiner window is now squeezed in, like a quarter the size it used to be. Maybe not related or another ticket.

image

schuelet (approved):

ahmadswaid:
@schuelet I think it's related to the layout setting in that workflow, could you please create another workflow to with a joiner component to check the behaviour is exist?

PR #778, opened 2021-02-25
fix unrecognizedPropertyException in ConversionUtils

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/245


schuelet (commented):
not sure if related, but if I pick 2 random models then the outport object is for some reason structured like a 4 model nested object and can't be read by the sim configurator. This doesn't happen with the salmonella - toymodel combination.

[Component_outport_bug.zip](https://github.com/SiLeBAT/FSK-Lab/files/6048729/Component_outport_bug.zip)

schuelet (approved):

PR #788, opened 2021-03-07
Fskdb view model re selection after saving

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/256


schuelet:
while this particular bug seems fixed, adding another connection between two models after reopening the view will bug the outport object again. Do you want to look at it in this ticket or in another?

image

ahmadswaid:
@schuelet: for the sake of issue simplisty let's create another ticket please.

schuelet (approved):

PR #791, opened 2021-03-09
fix simulations deletion in the Editor

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/183


schuelet:
Good work, Ahmad. The problem is, that if the editor is not opened at all, the object loses its simulations as well. So when I open the Editor View and save, everything works as intended, but if I then save the workflow and reset the editor, the simulations are gone again when executing the editor.
In short: open Editor > save > save workflow & restart> reset & re-execute editor > simulations are gone

ahmadswaid:
> Good work, Ahmad. The problem is, that if the editor is not opened at all, the object loses its simulations as well. So when I open the Editor View and save, everything works as intended, but if I then save the workflow and reset the editor, the simulations are gone again when executing the editor.
> In short: open Editor > save > save workflow & restart> reset & re-execute editor > simulations are gone

Thanks @schuelet, I missed that case, I add a one line to handle it.
but the still the whole code is nessesary to handle the other cases where the old code was removing all simulation in any case.
so I hope, now we cover all the use cases.

schuelet (approved):
Very good. I see that adding a parameter correctly removes the old simulations as well. Great.

PR #800, opened 2021-03-16
add exception if model script fails in scriptHandler runSnippet()

schuelet:
[error_model_script.zip](https://github.com/SiLeBAT/FSK-Lab/files/6150423/error_model_script.zip)

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/ScriptHandler.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/ScriptHandler.java
@@ -13,6 +13,7 @@ import org.knime.core.node.NodeLogger;
 import org.knime.core.util.FileUtil;
 import org.knime.python2.PythonVersion;
 import org.rosuda.REngine.REXPMismatchException;
+import de.bund.bfr.knime.fsklab.ModelScriptException;
 import de.bund.bfr.knime.fsklab.nodes.plot.ModelPlotter;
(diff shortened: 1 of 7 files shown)


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/ScriptHandler.java
```diff
@@ -103,8 +104,18 @@ public final void runSnippet(final FskPortObject fskObj, final FskSimulation sim
       
 
     exec.setProgress(0.75, "Run models script");
-    runScript(fskObj.getModel(), exec, false);
-
+    try {
+      runScript(fskObj.getModel(), exec, false);
```
miguelalba, line 108: The catch only captures RException and IOException but runScript also throws CanceledExecutionException, InterruptedException, REXPMismatchException. The latter 3 are not subclasses of RException or IOException so they are not caught. *runSnippet* throws Exception so if one of the latter 3 exceptions occur they are leaked by runSnippet.

miguelalba (changes requested):

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/ScriptHandler.java
schuelet: @miguelalba runscript() with "false" argument never throws a REXPMismatchException. And I don't know if we should catch CanceledExecution or InterruptedException here.

miguelalba (approved):

PR #803, opened 2021-03-16
fixing simulation scenarios generation

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/275
https://github.com/RakipInitiative/ModelRepository/issues/274


schuelet:
@ahmadswaid Thanks, it does work. However I am concerned that the first simulation is no longer called "defaultSimulation", not sure if it could break some node down the line. Do you think it is safe enough?

ahmadswaid:
@schuelet:  I don't think unsafe, but I will not take the risk. the simulation generated of "defaultsimulation"s is "defaultsimulation" from now on.

ahmadswaid:
@schuelet: Do you think you can reolve this PR?

schuelet (approved):
@ahmadswaid apologies, the PR showed no update on the "Pull requests" main page.

PR #804, opened 2021-03-16
fix https://github.com/RakipInitiative/ModelRepository/issues/276

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/276

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/reader/ReaderNodeModel.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/reader/ReaderNodeModel.java
@@ -102,7 +102,7 @@ public class ReaderNodeModel extends NoInternalsModel {
     } catch (IOException e) {
       e.printStackTrace();
     }
-    File fskxFile = new File(getDownloadedFileName());
+    File fskxFile = new File(getDownloadedFilePath());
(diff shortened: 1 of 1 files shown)


schuelet (changes requested):
Doesn't seem to work on Windows. Used the joiner_doesnt_add_simulations workflow.
`ERROR` FSKX Reader          0:35       Execute failed: Illegal char <:> at index 74: C:/Users/thsch/AppData/Local/Temp/knime_joiner_doesnt_a47845/FSKX Reader 0:35.fskx`

Also, getDownloadedFileName() seems unsafe, if the Reader is wrapped in a componend or metanode, it might screw up the path. Not sure though. I am also not sure if the File.separator would make sharing workflows between different systems impossible, since the environment manager looks for the path, but I could live with that for now.

schuelet:
The colon (":") was the problem, adding .replaceAll(":", "_") seems to solve the issue. 
However, 
- upon saving and reloading the wf, the temp archive is gone.
- wrapping in a metanode does not make the Runner fail, so that is not an issue after all.

schuelet:
I propose to just save the archive in the just in the workflow folder, maybe take the model id to name it, that would remove the chance to have duplicates if other reader nodes attempt to download the same model. It would also remove the issue (i hope) with sharing the wf across different OS.

ahmadswaid:
@schuelet can't use model ID as a name since we don't have it yet before downloading.

schuelet (approved):

PR #805, opened 2021-03-17
Fixes vocabularies in editor node

miguelalba:
The editor node (FSK lib) is updated to use the internal KNIME requests.

I changed a lot here. First the responses do not need to contain a wrapper (FskmlObject) but can be string arrays since we only use the names of the entries in the vocabularies. If we ever need the full entries we would have to use the wrapper.

Secondly I could not use the vocabularies from the editor even when calling from the library. In the end I had to patch the InputForm and SelectForm to make the requests via knimeService. For InputForm(s) I added a second addControlledVocabulary function that takes only an input and the vocabulary name. The other function is not modified. For SelectForm I just include the requests code inside _create. I commented the previous code. Please check if I broke functionality of other nodes (joiner?) since I am changing code of the library.

Regarding the requests, it was not needed to add a conditional module to load with require. Since the call is small it can be included in the method of loadConditionally. (lazyload.js gone)

Finally I also changed the dependencies. I put the knimeService libraries on top, before our library, so knimeService is available.


ahmadswaid:
Hi @miguelalba: 
- As I understood the PR is about providing a solution for controlled vocabullary problem in remote case, unfortunately the solution isn't working when I deploy a a workflow to a knime server having your solution's update site installed on my knime server.
- I think the changes in the JS libs order has broken something for the Editor in the server mode, now it doesn't open as it's loaded for the first time in the browser from the server.

I installed the local build of FSKLab on Knime Server 4.12 with 4.3 executor/ which always used to work.

[JoinerComponent.knwf.zip](https://github.com/SiLeBAT/FSK-Lab/files/6160188/JoinerComponent.knwf.zip)

miguelalba:
Please ignore this PR. Thomas reported errors with multiple views. This is replaced with https://github.com/SiLeBAT/FSK-Lab/pull/806

PR #807, opened 2021-03-18
model delivery bug to the joiner in Component mode

ahmadswaid:


schuelet:
@ahmadswaid Hi Ahmad, what was the issue here? Is this related to a ticket? Thanks.

ahmadswaid:
> @ahmadswaid Hi Ahmad, what was the issue here? Is this related to a ticket? Thanks.

 Hi @schuelet:
There is no ticket related to.
In the joiner component:
Select model for joining -> edit it and change somthing in the metadata like the model name or parameter's one-> the issue is that the changes are not delivered to the joiner to updated the selected model.

schuelet (approved):

PR #810, opened 2021-03-19
fix https://github.com/RakipInitiative/ModelRepository/issues/283

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/283
unfortunately, I can't reproduce the bug, but during testing I found a bug in editing the first selected model.
@schuelet could you please check if the problem still exist.


schuelet:
@ahmadswaid The problem is still there. try to edit 3 models, then select them for the Joiner and you will see a duplicate.

ahmadswaid:
@schuelet there is a bug in Knimeservice 4.0 event broadcasting, which is fix later in 4.2.
I add a fix to be compatible with both versions.
please check again!

schuelet (approved):

PR #811, opened 2021-03-19
fix shared setting between FSKDB view nodes

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/284


miguelalba (approved):
It is working for me.

Issue #820, opened 2021-03-24
FSK Editor JS UI is not working for long drop down list


mfilter:
Please check the FSK Editor in this WF: knime://BfR-extern/FSKX_Testing/FSK-Lab_2.0/broken_workflows/Test_MF

For example list drop down list for "language written in" or "licences" is not visible to its full extend, but the user has no chance to "scroll down" the list as there is no scrollbar.
Also the "info button" for "language written in" and some other metadata fields are not working anymore, e.g. design type, measurement type, fitting procedure ..., 

Further: the usability of the view for the "Parameter" is sub-optimal, as one can only look at the columns to the rights after scrolling down to the end of the table. The scroll bar shoud be always visible, if possible

schuelet:
fixed.

PR #821, opened 2021-03-24
Fsk editor js UI is not working for long drop down list

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/820


schuelet (approved):
@ahmadswaid works very well, great job. One question, what's with the hundreds of styles.css changes? Did you alter the look or is that just the way it was compiled?

ahmadswaid:
> @ahmadswaid works very well, great job. One question, what's with the hundreds of styles.css changes? Did you alter the look or is that just the way it was compiled?

regarding the style.css was not part of the grunt job so it was never changed and this was a bug if you want to change something related to the styling, I add it to the ignore so the will not be part of the future commits.

PR #829, opened 2021-03-28
fix a bug in editing reference dialog

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/296

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/editor/editor.js
+++ b/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/editor/editor.js
@@ -36,32 +36,12 @@ fskeditorjs = function () {
     //}
   }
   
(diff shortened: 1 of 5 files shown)


schuelet (changes requested):
@ great job. I noticed an issue though: 
1. the checkbox "is reference" seems to be applied for the entire array, i.e. the last checkbox value is used for all entries. If I edit 2 references an give "true" to reference 1 and "false" to reference 2, then both are "false" after saving.
2. Please remove the function removeUndefinedFromJson, since it seems no longer be necessary and might mess up saving the array. Thanks.

schuelet (changes requested):
@ahmadswaid I still have the problem that you can't edit more than one reference. Is this property cursed?
editor_original

editor_modified

portobject_modified

ahmadswaid:
@schuelet, sorry for going back and forth, It's seems that using `splice` then `push` functions a new row doesn't work as promised.
I use only `splice` to remove and add an item in one statment and that worked fine for reference and parameter and other list.

schuelet (approved):
@ahmadswaid great job, it works

PR #832, opened 2021-03-31
Db view make some of the landingpage elements configurable

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/306
- Sorry for the huge PR but it can't be divided in parts otherwise it can't be tested.
- I used the following workflow with local repository to simulate the web repository workflow.
[JoinerComponent.knwf.zip](https://github.com/SiLeBAT/FSK-Lab/files/6240098/JoinerComponent.knwf.zip)


schuelet:
@ahmadswaid very impressive. I have a view questions:
- how does the dbview load the simulations? They are not part of the metadata of a model and can't just be added to a knime table.
- how does the download button work if the path points to a local folder?
- where do I need to put the execution time/ upload date information so that it is displayed in the table view?

ahmadswaid:
> @ahmadswaid very impressive. I have a view questions:
> 
>     * how does the dbview load the simulations? They are not part of the metadata of a model and can't just be added to a knime table.
> 
>     * how does the download button work if the path points to a local folder?
> 
>     * where do I need to put the execution time/ upload date information so that it is displayed in the table view?
> 
>     * and since we're at it, how can we make the search function work for local paths if at all?
>       Because non of that is trivial, I suspect we need to update the landingpage backend and use that on the server if it allows that. What do you think?

@schuelet all of these resources (the plots, simulations and the scripts) are fetched from the landingpage backed and also the running after selecting the simulation is happening on that server.
but for the execution time and upload date they should be added to the columns of the generated table out of the loop I simulate that by adding rendom values considring that you also will get it from local file if there is an online service for that I can easly change that behaviour.
The filtering is working in the local html table but for the search, It uses the online service for that and if we have the backend service has the same database as the generated knime table then I think it will work as good but to be honest I didn't tested it thoroughly.

llavall:
Hi all, I was wondering if we could have a schemata how all those parts work together? Is that component something which will only be used on our server or should it work everywhere?

ahmadswaid:
@schuelet I have added the support to the web repository with online backend and made the execution configurable.
please have another look.

ahmadswaid:
@llavall it's client/server base solution and because of that @schuelet added this ticket: https://github.com/RakipInitiative/ModelRepository/issues/307

schuelet:
@ahmadswaid the joiner doesn't seem to work if I have the buttons active. Selecting models don't show up in the joiner view. Having no buttons active works.

ahmadswaid:
@schuelet, sorry this was by design, I didn't know that we have to support that case.
Anyway I added the required code to support that.

schuelet:
@ahmadswaid if an optional button is set to true, the edit button is no longer there. At least the header option should work when the DBView node is used for joining, and if possible the download button as well.

ahmadswaid:
@schuelet: I pushed the required changes:
- edit button always available in component mode
- header and download button if option is selected
please have another look.

schuelet (approved):
@ahmadswaid thanks, very good work.

mfilter:
I think we need a proper node description now as well. Specifically it is important to clarify if this node can handle models that are in local folders. If not we should provide a pointer to a workflow that illustrate how this can be enabled. This workflow could even be hosted on the KNIME community hub.

PR #835, opened 2021-04-08
removed validation for email in editor

schuelet:
At the request of @mfilter I removed the validation of emails in the editor.
@ahmadswaid please give this a thought if it breaks anything down the line, in my tests (runner -> writer->reader) it worked fine.

### DIFF
--- a/de.bund.bfr.knime.js/src/js/app/app.editable.mt.InputForm.js
+++ b/de.bund.bfr.knime.js/src/js/app/app.editable.mt.InputForm.js
@@ -266,20 +266,19 @@
 				}
 			
             }else {
-				
-				isValid = O.input.val() ? true : false;
-				if(O.type === "email"){
-					isValid = true;	
-				}
+                isValid = O.input.val() ? true : false;
                 if(!isValid){
                     O.input.$validationContainer.text("required");
                 }
-                // check if mail has correct structure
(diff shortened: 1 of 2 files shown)


de.bund.bfr.knime.js/src/js/app/app.editable.mt.InputForm.js
```diff
@@ -266,15 +266,19 @@
...
+				}
                 if(!isValid){
                     O.input.$validationContainer.text("required");
                 }
                 // check if mail has correct structure
-                if(isValid && O.type === "email"){
-                    isValid= /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/.test(O.input.val());
-                    O.input.$validationContainer.text("Not a valid email value");
-                }
+                //if(isValid && O.type === "email"){
```
ahmadswaid, line 278: I don't think we should cancel the validation if the user provide some text.
As I understood the request was about making it optional and that can be achieved by relaxing the schema in editor_data_1.0.0.js and add the required code to validate the email text if provided.

schuelet:
@ahmadswaid very well, I added another commit with your suggestions.

ahmadswaid (approved):

PR #837, opened 2021-04-08
Manipulate joiner svg

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/309
I was able to remove all the annotation related to ports tooltip or select box (this was asked by @mfilter a while ago)
now I don't see the problem anymore.
I removed also the arrow heads in the SVG (if you see thart I keep only the target arrowhead or any thing else, just mention it)

I uploaded a simple test workflow here: knime://knime.mountpoint/FSKX_Testing/FSK-Lab_2.0/birt_report

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/joiner/joiner.js
+++ b/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/joiner/joiner.js
@@ -372,6 +372,8 @@ joiner = function () {
     }
     d3.select("svg").selectAll(".link-tools").remove();
     d3.select("svg").selectAll(".marker-arrowhead-group").remove();
+    d3.select("svg").selectAll("title").remove();
+    d3.select("svg").selectAll("foreignObject").remove();
     _value.svgRepresentation = this.getSVG();
     if(_graph)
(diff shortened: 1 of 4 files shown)


schuelet:
@ahmadswaid very well, I think it almost works. Either this filtering doesn't apply to the Joiner Component, or some more filtering needs to be done: 
knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/birt_report_thomas

The upper svg is from the isolated joiner, the lower is from joining the first 2 models of the landingpage
image

schuelet:
just in case it is useful, here  is the image file: 
[joiner_image.zip](https://github.com/SiLeBAT/FSK-Lab/files/6286506/joiner_image.zip)

ahmadswaid:
@schuelet: at the end I was able to reproduce it on windows by installing a new knime, It's seems that knime on windows doesn't support by default UTF-8 Charset.
One of the selected models for joining in your example contains some special charachters like `µ`, even special chars (like line break) will not be recognized if the charset of knime isn't set correctly to handle these cases.

Please set: -Dfile.encoding=UTF-8 in your knime.ini to solve the issue.

mfilter:
I think, we should report this to KNIME.
One other point: I would really prefer if we could keep the target arrowheads.

schuelet (approved):

Issue #838, opened 2021-04-10
Edit Parameter view opened from the new editor does not display the correct information


mfilter:
Got to https://knime.bfr.berlin/knime/webportal/space/FSK-Web/JoinerComponent_Petra_modified
select any of the models and click the "edit" symbol
in the editor view go to "Model math -> Parameter
edit any of the parameters in the model and check the information on Data type and Classification.
In the pop up window many parameters are wrongly labeled as "Constant" even though they are declared as Input or Output.
The same is true for the data type that is displayed. 
I did not check other fields, but this  must not happen.

schuelet:
@ahmadswaid to clarify, the issue is, that on opening the modal view of "edit parameter", the properties "classification" and "data type" are set to their default values (or first values in the list), i.e. for "data type", it is always set to INTEGER. This does not seem to occur on other fields using controlled vocabulary .

schuelet:
fixed as of https://app.zenhub.com/workspace/o/silebat/fsk-lab/issues/843 and  https://github.com/SiLeBAT/FSK-Lab/pull/846

Issue #357, opened 2019-08-12
Editor JS: model metainformation from drop-down list is not correctly displayed for existing models


schuelet:
General >Model Category>Model Class list in drop down menu is incomplete

> 3.	By the way the Model Class select list does not display all the model categories. 
![image](https://user-images.githubusercontent.com/19906968/63438257-4b279080-c42c-11e9-9922-f785593f6d4e.png)

mfilter:
The original issue is probably solved, but the Editor node dose not display the correct information for model class when it is linked to an existing model. At least not for the model class, where the first entry of the drop-down list is displayed instead of the correct value. See 
knime://BfR-extern/FSKX_Testing/FSK-Lab_2.0/broken_workflows/testCampy  
General information -> Model caegory -> Model class: here "[Data]" is shown, but it should be "Dose-response model".
I think this is related with similar observations in other tickets.

PR #843, opened 2021-04-11
update selected value

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/357
https://github.com/SiLeBAT/FSK-Lab/issues/838
one finding here is that the list of parameter classification values have non-capital letters which leads to have the value not selected by editing, a ticket has to be created to fix the value in the controlled vocabularies if on is not exist yet.


schuelet:
@ahmadswaid  on opening a model with a parameter, the parameter classification select is emtpy and on saving there is no longer a classification set even if there was one before: 
image

image

schuelet:
@ahmadswaid also, I just noticed, the header says "add Parameter", maybe we should just change that to "Parameter" since it uses the same modal as the "add parameter" button.

schuelet (approved):

PR #844, opened 2021-04-12
retrieve css for fskdbview pre tag

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/313


schuelet (approved):

ahmadswaid:
@schuelet @miguelalba: it's not enouph to have the approval to merge.

mfilter:
@schuelet : is the issue with the missing public key something that @miguelalba must take care of?

schuelet:
@ahmadswaid I switched the distribution to Xenial (which is a newer Ubuntu) and that seems to work.  I am waiting until it is done and then create a PR for you to approve, then the issue should be gone for now.

PR #845, opened 2021-04-13
switched to dist: Xenial

schuelet:
changed travis.yml to use another distribution of ubuntu


ahmadswaid (approved):

schuelet:
@llavall The platform where we build FSK-Lab is not relevant, it could just as well be windows or Mac. Ubuntu is what Travis provides and what is easiest to use.

Issue #360, opened 2019-08-12
Editor JS: mandatory fields are not enforced


schuelet:
mandatory fields are not enforced (ref. Esti 5.)

> In Model Math tab if I add a new parameter and I do not complete all the required fields, I see the warning indicating how many required properties are missing, but if I close the node and save the changes no error is shown. Actually I can still run the model but if then I try to use the FSKX Writer node I get this error: Execute failed: (“Null pointer exception”): null. I remember that in previous versions it used to show on the bottom of the Editor node a warning saying that something was wrong when some information was missing.

![image](https://user-images.githubusercontent.com/19906968/63446504-9e084480-c43a-11e9-897c-832a9d01d313.png)

schuelet:
This is still the case and I suggest we keep it simple and try to add a warning if so-called mandatory fields are not filled out. Saving the editor should always be possible, in case someone wants to continue working on it later.

mfilter:
agree

schuelet:
@ahmadswaid : java script side: 
- window pop up if user wants to close editor with missing mandatory fields. Window includes all information about missing poperties.
- transfer information to java side (or prepare transfer)

@miguelalba 
- get warning messages from javascript to java side
- warning (node) if editor is saved with missing mandatory fields
-

schuelet:
fixed as of https://app.zenhub.com/workspace/o/silebat/fsk-lab/issues/849

PR #849, opened 2021-04-15
Editor validation

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/360:
- both javascript and java side are implemented.
- JSON schema is created from YAML schema in fsk lab (this can be automated later).
- no way to call javascript function to confirm validation before showing save message.


schuelet (approved):
seems to work. Great job.

Issue #860, opened 2021-04-19
Marcel Editor 1: Editing entries in Scope not working


Thatsreallycool:
Adding entries via Editor in "Scope->Product", "Scope->Hazard", "Scope->Population group" works fine, but editing entries once saved do not lead to the Editing pop-up window.

Tested on existing models and a simple Editor with test entries with the same outcome

![editor](https://user-images.githubusercontent.com/37627583/115193340-5133b000-a0ec-11eb-80d6-ca45b66d5155.png)

schuelet:
Here is a workflow showcasing the issue. It seems some missing values cause a null pointer exception.
knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/editor_bug_editing_in_scope

knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/editor_bug_editing_in_scope

ahmadswaid:
@filter I have already solved it in https://github.com/SiLeBAT/FSK-Lab/pull/863, I was waiting to @schuelet to close the current ticket

mfilter:
Marcel just tested it and for him it did not work. I assign Marcel now as well.

mfilter:
Maybe you could have a web meeting to clarify

ahmadswaid:
@mfilter I don't if @Thatsreallycool  tested it on a build which contains the fix.
anyway trying to contact him.

mfilter:
@Thatsreallycool :
please confirm that this issue is solved with the latest built

Thatsreallycool:
tested this on 
https://gitlab.bfr.berlin/silebat/repositories/raw/master/fsklab/

with this build this bug is not encountered. 
Before closing this ticket, i would suggest to wait until this can be tested on "test".

mfilter:
seems to work now. I close this ticket. Great!

PR #863, opened 2021-04-20
fix: editor doesn't enable editing newly added rows in any table

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/860

### DIFF
--- a/de.bund.bfr.knime.js/src/js/app/app.editable.mt.Dialog.js
+++ b/de.bund.bfr.knime.js/src/js/app/app.editable.mt.Dialog.js
@@ -112,7 +112,11 @@
                 let data = {};
                 for (const inputId in O.inputs) {
                     let currentInput = O.inputs[inputId];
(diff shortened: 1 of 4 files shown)


schuelet:
I am using the model knime://knime.mountpoint/FSK-Web/newfiles/202009031723_DRMclassicalONLY.fskx

Reading this model, I can't add a new product. Hazard works as intended. Maybe some element is missing in this particular model?

mfilter:
I think this is a feature, not a bug.
Dose response models do not have products - see the Generic Metadata Schema: https://foodrisklabs.bfr.bund.de/wp-content/uploads/2020/11/Metadata_-Master_Table_V1.04.xlsx
HTH;
MF

ahmadswaid:
@mfilter thank you, unfortunatly that feature is not impemented. maybe a ticket has to be created for that case, if one isnot exist yet.
@schuelet the issue was related to bug in old commits. thanks fr reporting.
anyway I have fixed them all and try my best to test adding, editing new rows existed or newly created rows in all modals belong to the editor.
Please have another review. and thank you in advanced.

schuelet (approved):
looks good.

PR #865, opened 2021-04-22
default svg generation if vis script doesn't generate one

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/206
this is how I see thing has to be implemented if no svg generated out of visualization script or in case an error occurred there.
I'm open to any further changes.


schuelet:
@ahmadswaid I am against catching the default exception. If something in the execution failed, the node should fail. I propose to check if no viz script is given or empty, then apply the svg code. Thanks.

schuelet:
@ahmadswaid If the execution throws an error because the viz script is empty or not there, we could also think about just making a default viz script and apply it to the model. Whatever works for you though.

ahmadswaid:
> @ahmadswaid I am against catching the default exception. If something in the execution failed, the node should fail. I propose to check if no viz script is given or empty, then apply the svg code. Thanks.
 @schuelet I will take this way and check if vis not available   then generate the default SVG.

schuelet (approved):

Issue #398, opened 2020-04-16
Add validation alerts to JS Simulation configurator node


miguelalba:
The validation of the simulation configurator node is performed for each parameter but in the case of models with many parameters most are hidden. An alert will be handy when the validation fails.
The alert should list the parameters with errors. See:
- https://getbootstrap.com/docs/4.0/components/modal/
- https://getbootstrap.com/docs/4.0/components/alerts/

PR #869, opened 2021-04-25
add global validation message area

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/398#event-4577356426


schuelet:
@ahmadswaid when is the alert supposed to be triggered? If I leave fields empty or wirte text instead of number values, I am not getting any messages.

ahmadswaid:
image
@schuelet if you leave the fields empty in the simulator configuration on save, you should see the global error messages.

schuelet (approved):

Issue #489, opened 2020-07-21
esti simulator 5: validate if parameter values have correct data type and form


schuelet:
e.g:  if field is a number -> string value should be forbidden

PR #871, opened 2021-04-25
validate integer values

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/489
@schuelet:
for interger and double only numeric values are allowed (already provided via the component) but I added Integer validation.
for other data types (in my point of view ) this will rais some limitation to the user freedom in providing the values. String values could be a literal string starting and end with `"` or could be an assignment to another string variable. there a lot of ways to define a matrix vector or boolean.

### DIFF
--- a/de.bund.bfr.knime.js/src/js/app/app.simulation.js
+++ b/de.bund.bfr.knime.js/src/js/app/app.simulation.js
@@ -952,7 +952,13 @@ class APPSimulation {
 				// check range for integers and doubles
 				if ( field.param.dataType.toLowerCase() === 'integer' 
 					|| field.param.dataType.toLowerCase() === 'double' ) {
-
+                    
+                    if ( field.param.dataType.toLowerCase === 'integer' && fieldValue.indexOf('.') > 0 ){
+                        return {
+                                input   : field.input,
+                                msg     : 'Invalid integer value'
+                            };
+                    }
 					if ( field.param.minValue && parseFloat( field.param.minValue ) > fieldValue ) {
 						return {
 							input  	: field.input,


schuelet:
@ahmadswaid I am not so sure about that ticket anymore. Valid assignments of numeric parameters could be:
x <- 20
y <- sqrt(x)

And at Javascript runtime, we can not validate that. I propose to let this one go.

ahmadswaid:
@schuelet numeric values are controlled with this component 
image
String literals are not allowed

schuelet:
@ahmadswaid That's what I mean, what if the parameter is a double but is declared as a reference to a previous parameter:
my_double1 = 0.2
my_double2 = sqrt(my_double1)

ahmadswaid:
> @ahmadswaid That's what I mean, what if the parameter is a double but is declared as a reference to a previous parameter:
> my_double1 = 0.2
> my_double2 = sqrt(my_double1)

there is no way for numeric variables to refrence another variables  with the current view only literal numeric values are allowed

schuelet:
@ahmadswaid Here is the new_Duarte_V2 Model:

image

ahmadswaid:
> @ahmadswaid Here is the new_Duarte_V2 Model:
> 
> image

@schuelet  try to create another simulation and change numeric values

schuelet:
@ahmadswaid That still works. I only changed the first parameter and it still saves the view.

schuelet:
@ahmadswaid but editing mu_start for example resets the view to 0
image

ahmadswaid:
@schuelet yes this what I meant

schuelet:
@mfilter This is a dilemma, we can't have the nice sliders and +/- buttons on numeric values if modellers decide to use references to other parameters. This would need to be checked on model creation or the parameter type needs to be set to STRING in the Simulation Configuration to make editing still workable.

mfilter:
Yes, I agree this is a dilemma.
I tend towards the proposal from Ahmad, and force all those inputs to use the string data type where a command is provided.
We can discuss this tomorrow. 
Hth,
Mf

>>> Thomas ***@***.***> 25.04.21 13.13 Uhr >>>
@mfilter This is a dilemma, we can't have the nice sliders and +/- buttons on numeric values if modellers decide to use references to other parameters. This would need to be checked on model creation or the parameter type needs to be set to STRING in the Simulation Configuration to make editing still workable.

-- 
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub:
https://github.com/SiLeBAT/FSK-Lab/pull/871#issuecomment-826306460
/// Wissenschaft im Dienst des Menschen ///
/// Science in the service of humanity ///

schuelet (approved):

Issue #490, opened 2020-07-21
Esti Runner 6 : enable auto-open image in Runner node


schuelet:

schuelet:
done

PR #873, opened 2021-04-25
enable auto-open image in Runner

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/490#event-4577257212

@schuelet : you can test it by clicking the new item in the right click menu:
image
I added the batik libraries to make use of the generated svg in the popup panel

thanks


schuelet (approved):
@ahmadswaid cool!

PR #876, opened 2021-04-26
escape HTML special Chars

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/320


schuelet (approved):
yes, works. Thanks very much.

ahmadswaid:
there is a failing travis job which prevents merging

PR #879, opened 2021-04-28
Resources

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/104
@schuelet: 
- test has to be done on server.
- resources will be hosted in a temp folder belongs to the current executed editor 
- all temp folders will be located in a folder called tempResources in Knime server mount point.
- the editor will move these resource files to the working directory of the fskObject, then delete the temp folder with its content.

Thanks, and sorry for the big PR.

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/editor/editor.js
+++ b/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/editor/editor.js
@@ -101,9 +101,6 @@ fskeditorjs = function () {
         // this folder will be removed after coping all the content inside
         // to the fsk object working directory.
         var anotherxhttp = new XMLHttpRequest();
-        anotherxhttp.onreadystatechange = function(response) {
-           console.log(response);
-        };
         parentResourcesFolder = "knime://knime.mountpoint/tempResources/jsEditorTempFolder"
                 + timeStampInMs;
         anotherxhttp.open("put", server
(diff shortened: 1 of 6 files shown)


llavall:
> all temp folders will be located in a folder called tempResources in Knime server mount point.

@ahmadswaid  What is the advantage to create a temp folder in the KNIME Server Mountpoints instead in the temp folder of the server? Where is that folder located? I thought we want to store such stuff inside of the job now, in order to be automatically be removed if the hob is removed?

ahmadswaid:
@llavall it will be removed atomatically even before by the moving to the next node in the workflow and will not live until the job is killed.
also this is managed by knime REST API/ the server is controlling where to create such resource.

llavall:
@ahmadswaid and what is the advantage over the server /tmp folder or the temp folder inside of the job? It sound like all over the server are such folders created then.

schuelet (approved):
@ahmadswaid As far as I can see, it works, pretty impressive job. 
I tested:
- creating a model with .r, .csv and .xlsx resources
- adding a new file to an existing model
- adding a source file that the model uses in the Runner

 This works so much better than the current local environment manager, I wish this was possible on local as well, but that is an issue for another time.

Thatsreallycool:
Creating a model and then using the Environment Tab leads to deleting all previously entered data, if not stored as default.

Model that is supposed to be loaded from the reader is also being ignored by the editor after using the Environment Tab.

example workflow provided in 
knime://BfR-extern/FSKX_Testing/FSK-Lab_2.0/broken_workflows/Marcel_testing_ressource_tab

ahmadswaid:
@Thatsreallycool OK, now I understand, but just not to mix things, here is technical ticket for merging the `server side resource tab feature`.
if you have the task to test the new feature then you should create and run a workflow containing the editor on the server there you find a tab called `Resources` which you can test.
anyway I will take care of the environment tab issue.

Issue #875, opened 2021-04-26
Marcel Editor 2: edited parameters are not stored in fsk-object at outport


Thatsreallycool:
update site: fsktest - https://dl.bintray.com/silebat/test2

When editing the list of parameters via Editor, adding new parameters or renaming existing parameters does not update the outgoing fsk-object. A Runner runs the model with the old set of parameters and values

uploaded a test workflow for this behaviour
knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/editor_renaming_parameters

schuelet:
@Thatsreallycool In the current version, this does not happen. Try using the new BfR update site:
https://gitlab.bfr.berlin/silebat/repositories/raw/master/fsklab/

Here is the modified workflow (reset may be necessary):
knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/fixed_broken_workflows/editor_renaming_parameters_TS

Thatsreallycool:
Tested it, and it works fine with this build. Thanks!

Thatsreallycool:
update site: https://gitlab.bfr.berlin/silebat/repositories/raw/master/fsklab/

With the newest update, this bug occured again. 
Added new parameters via Editor, Runner cannot "see" them. Saved Model with Writer. 
Then used the reader to load the file. Still, the reader is not able to load the new set of parameters. But when the Editor is opened, all new parameters are present.

Workflow provided in:
knime://BfR-extern/FSKX_Testing/FSK-Lab_2.0/broken_workflows/Marcel_editor_adding_new_parameters

mfilter:
@Thatsreallycool : could you please explain which parameter you added that are not visible in the Reader? That would help others to reproduce your findings.

mfilter:
I can confirm the findings. A side-note: many of the new parameters are defined as "Constant"s. So I would not expect those to be listed in a the simulations, but all the others. I assume one needs to trigger the creation of the simulation scenarios every time the editor is closed.

Thatsreallycool:
> 
> 
> I can confirm the findings. A side-note: many of the new parameters are defined as "Constant"s. So I would not expect those to be listed in a the simulations, but all the others. I assume one needs to trigger the creation of the simulation scenarios every time the editor is closed.

Just a quick side note for clarity: According to the errorlog of the runner,  both newly added parameter types (const and input) are not found, when applying the runner.

[...]

PR #880, opened 2021-04-28
retrieve originalMetadata from viewRep

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/issues/875


schuelet:
@ahmadswaid I think the problem is that newly created parameters are not saved in a simulation. Is it possible to create a new default simulation if a new input or constant parameter is added? Right now, when using the editor, this is what the Simulation Configurator shows: 

Screenshot 2021-04-29 092304

schuelet (approved):

PR #884, opened 2021-05-01
Edited parameters are not stored in fsk object

ahmadswaid:
https://github.com/SiLeBAT/FSK-Lab/pull/879#issuecomment-829969499
https://github.com/SiLeBAT/FSK-Lab/issues/875

@schuelet this PR solved the issues:
- the one reported by @Thatsreallycool which is related to a reset on dialog saving.
- regenerate simulation for non-connected Editor.
- setting reset managment, that is an issue you notice when opening > saving > interactive view > save then problem will start.

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/editor/editor.js
+++ b/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/editor/editor.js
@@ -318,7 +318,7 @@ fskeditorjs = function () {
       readme: _readmeCodeMirror ? _readmeCodeMirror.getValue() : _metadata.readme,
       resourceFiles: _val.resourceFiles, // TODO: get actual resource files from editor
       serverName: _val.serverName, // TODO: get actual serverName from editor?
-      isCompleted: true,
+      completed: true,
       validationErrors: [],
       modelType: _metadata.modelType
     };
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/editor/FSKEditorJSConfig.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/editor/FSKEditorJSConfig.java
@@ -42,6 +42,7 @@ class FSKEditorJSConfig {
   private static final String ENVIRONMENT = "environment";
   private static final String SERVER_NAME = "serverName";
   private static final String COMPLETED = "completed";
+  private static final String REPRESET = "representationResetReuired";
   private static final String ERRORS = "errors";
   private static final String MODEL_TYPE = "modelType";
 
@@ -52,6 +53,7 @@ class FSKEditorJSConfig {
   private EnvironmentManager m_environment;
   private String m_serverName;
   private boolean m_isCompleted;
+  private boolean representationResetReuired = true;
   private String[] m_validationErrors;
   private String m_modelType;
 
@@ -148,6 +150,13 @@ class FSKEditorJSConfig {
   public void setModelType(String modelType) {
     m_modelType = modelType;
   }
+  public boolean isRepresentationResetReuired() {
+    return representationResetReuired;
+  }
+
+  public void setRepresentationResetReuired(boolean representationResetReuired) {
+    this.representationResetReuired = representationResetReuired;
+  }
 
   /**
    * Saves current parameters to settings object.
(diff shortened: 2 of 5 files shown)


schuelet:
@ahmadswaid using Matthias' workflow knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/Marcel_editor_adding_new_parameters_MF, this is the current behavior:
On opening Editor:
- if only input parameter is changed -> new parameters show up in Simulation Configurator and Runner settings, but simulations are removed (this is by design to avoid conflicts) 
- if only output parameter is changed -> new parameters don't show up in Runner settings and are without values in Simulation Configurator.

@mfilter These issue with this workflow is, that the model has been written already in a faulty state, i.e. with a set of simulations that have parameters without values. If you just slightly change an input parameter (e.g. add a space to its description) then writing and reading the model works with all new parameters attached. The other simulations are gone, but this is expected in the current implementation and I don't know if it is wise to try and keep simulations if the parameter metadata (for input/constant parameters) changes.

ahmadswaid:
@schuelet this is the expected behaviour because like you said if the state of the editor is saved(faulty) and if we don't change any input parameter then nothing will be changed in the simulation.

schuelet (approved):

mfilter:
Not sure if the procedure @schuelet proposed as "expected" matches the expectations of the users.
If you have already worked for days on a model, added specific simulations and you read in this model into the Editor to make an update you would not expect that old simulations are deleted, just because you added a new input parameter.
Would it be possible to just add the newly introduced input parameter to all existing simulations with the default value? In cases where there were changes on existing input parameters I would also think, that deleting simulations is not ideal. I think, it would be best to just "refresh" the default simulation (in case the modified parameter id was not changed) or delete the corresponing parameter entry in all simulations in cases where an existing input id was re-named. I know this is getting tricky. 
For sure, we should properly describe the behaviour in the node description.

mfilter:
here another WF illustrating my use case that illustrates why it is useful to enable changes to model parameters: knime://knime.mountpoint/FSKX_Testing/FSK-Lab_2.0/broken_workflows/gropin_join
Here I change two parameters from Input to Constants, as this makes much more sence. 
Even deleting an Input parameter would not be reflected on the FSK outport port (Java side) of the Editor. A deleted parameter would not be visible in the next Simulation configurator, but a Runner would still use this parameter.

So, currently there is no way for a user to have changes in a parameter securely reflected in the next node, which is a problem.

ahmadswaid:
> here another WF illustrating my use case that illustrates why it is useful to enable changes to model parameters: knime://knime.mountpoint/FSKX_Testing/FSK-Lab_2.0/broken_workflows/gropin_join
> Here I change two parameters from Input to Constants, as this makes much more sence.
> Even deleting an Input parameter would not be reflected on the FSK outport port (Java side) of the Editor. A deleted parameter would not be visible in the next Simulation configurator, but a Runner would still use this parameter.
> 
> So, currently there is no way for a user to have changes in a parameter securely reflected in the next node, which is a problem.

@mfilter thanks, just add a fix (https://github.com/SiLeBAT/FSK-Lab/pull/888) for the parameters deletion issue.
what I don't understand is what is the expected behavior if someone change the param type from input into constant.

mfilter:
the expected behavior is that the parameter is deleted from the simulation, as constants can't be changed by the user. The same should happen if an Input is changed to Output (for whatever reason). In case a constant parameter is changed to Input it should show up in all simulations.

schuelet:
@mfilter Not to pile on to this discussion, but you said a long time ago you wanted to have constant parameters to show up in the Simulation Configurator (just not editable by the user), so they are part of the simulation in that sense.

mfilter:
OK, indeed I did'nt remember that. Then we do not have a problem. Please check that a change in the parameter type classification is saved and any "re-labelled" constant parameter becomes "static" . Thx

schuelet:
Checked it, it works. The only issue, for later maybe, is that constants have no special labeling in the SimConfig (there is no property for classification), so we might want to think about adding some visual indicator when we have time.

schuelet:
> Not sure if the procedure @schuelet proposed as "expected" matches the expectations of the users.
> If you have already worked for days on a model, added specific simulations and you read in this model into the Editor to make an update you would not expect that old simulations are deleted, just because you added a new input parameter.
> Would it be possible to just add the newly introduced input parameter to all existing simulations with the default value? In cases where there were changes on existing input parameters I would also think, that deleting simulations is not ideal. I think, it would be best to just "refresh" the default simulation (in case the modified parameter id was not changed) or delete the corresponing parameter entry in all simulations in cases where an existing input id was re-named. I know this is getting tricky.
> For sure, we should properly describe the behaviour in the node description.

@mfilter I will open a ticket so we keep that in mind. It is however not that easy, I think, since we could have a complex situation like this: Input Parameters x,y,z with several simulations:
- delete x
- rename y to x
- turn z into output parameter
You would need to log all changes in that order to apply to all simulations otherwise you might not have all information on the java side to deal with the question which parameter is which and what value it is now supposed to have.

mfilter:
agree

PR #885, opened 2021-05-01
enable editing numeric parameters that contain text

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/321

entering a string value will destroy the touchspin nature of the field and re-entering a numeric value will retreive it again


schuelet:
@ahmadswaid I will review this when the issue RakipInitiative/ModelRepository#323 with the Details Button is solved. I am worried, that this might introduce another issue with the DBView.

ahmadswaid:
@schuelet OK, but first you need to have both branches locally merged.

schuelet (approved):

PR #887, opened 2021-05-02
enable edit button only if editor available in component other button

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/324


schuelet:
@ahmadswaid thx for the quick work, but now the models don't show up in the Joiner area anymore.

ahmadswaid:
> @ahmadswaid thx for the quick work, but now the models don't show up in the Joiner area anymore.

@schuelet the joiner is working on my side, could be a wrong setting of the DBView/ could you please use new one.

ahmadswaid:
@schuelet you're right, your finding uncover a hidden bug/ local files were not read from local repository in all cases.

schuelet (approved):

PR #895, opened 2021-05-03
fix CombineArchive concurrent access issue

ahmadswaid:
@schuelet:
I did before this PR a stress test trying to access the same file from 200 reader nodes at the same time and that passed successfuly.
also run two copies of the huge reader workflow from two different knime instances, trying to access the same poor file hundreds of times concurrently from two different JVM and that passed successfuly.

I do see the CombineArchive issue only in the Reader if there other places please issue it to me.


schuelet:
@ahmadswaid looks very good, just a question: The try(combineArchive...) block doesn't catch Exceptions, is it guaranteed that the finally - block is executed and the in.getPath() removed from the registry?

schuelet:
@ahmadswaid This works great. The other place where CombineArchive is used is in the ArchivedEnvironmentManager. Should we open a separate ticket for that?

ahmadswaid:
> @ahmadswaid looks very good, just a question: The try(combineArchive...) block doesn't catch Exceptions, is it guaranteed that the finally - block is executed and the in.getPath() removed from the registry?
@schuelet yes it's garantied, The finally block will always executes last even if uncaught exception throw in try block.

ahmadswaid:
> ArchivedEnvironmentManager

weird why I didn't found that yesterday, I would prefer to do it here as it's relevant to the topic

ahmadswaid:
@schuelet implemented also in ArchivedEnvironmentManager, please have another review

schuelet (approved):

PR #904, opened 2021-05-08
keep and update simulation on adding/edit/delete of parameter

ahmadswaid:
from discussion of https://github.com/SiLeBAT/FSK-Lab/pull/884

- adding a new parameter will be added to all simulations.
- editing the value of a parameter in the model math will change the value of that param in the default simulation.
- delete a parameter will delete that param from all simulations.


schuelet:
@ahmadswaid It seems the editor, when only executing without opening the view, tries to add all parameters (including OUTPUT) to the simulation. See here, the output parameter is called "Value" and is apparently added to the simulation:
image


see here:
knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/add_delete_params_and_sims

ahmadswaid:
> @ahmadswaid It seems the editor, when only executing without opening the view, tries to add all parameters (including OUTPUT) to the simulation. See here, the output parameter is called "Value" and is apparently added to the simulation:
> image
> 
> see here:
> knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/add_delete_params_and_sims

@schuelet good catch, I fix it please re-check

schuelet (approved):
very good. Test workflow added to external server.

PR #909, opened 2021-05-09
forbid constant parameters ovveride

ahmadswaid:
having input_conso as CONSTANT.
and It's not allowed to override this kind of parameters.
image


schuelet:
@ahmadswaid I did merge development into the master branch, so I don't know if you need to reupload the PR. What would be the proper procedure? Delete development and create a new one?

ahmadswaid:
> @ahmadswaid I did merge development into the master branch, so I don't know if you need to reupload the PR. What would be the proper procedure? Delete development and create a new one?

I think we can continue working with the PRs as usual. no need to delete or reupload.
just approve the current PR if it doesn't have any issue then I will merge it into Development. and if you want you can then merge development into master so that you have both synchronized.

schuelet (approved):
nice work.

PR #910, opened 2021-05-14
R libraries management

ahmadswaid:
@schuelet: please allow me take part of your email to list what I implemented in this PR:
1. FSK-Lab uses R 3.4.4 bundle: 
    we use only the .fsk folder , and completely ignore any user folders when scanning for installed packages
2. FSK-Lab uses user defined R:
    we don't use the .fsk folder and just use the default library folder, the user has installed. Auto-install of packages goes then   only to the default library folder( the first value returned by calling .libpath() function)
3- if the user decided to change the R setting to use his own R binary rather than BfR R bundle then the LibRegistry instance will be forced to be refreshed.
4. in case the user is using R 3.4.4 bundle and triggers a run then the .RProfile will be configured temporarily during the run to reference the .fsk folder and the original configuration will be restored after finishing the current runner if there is no other runner is still working.

Thanks.

### DIFF
--- a/de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/RprofileManager.java
+++ b/de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/RprofileManager.java
@@ -1,0 +1,40 @@
+package de.bund.bfr.knime.fsklab.r.client;
+
+import java.io.IOException;
+import java.util.concurrent.locks.ReentrantLock;
+import de.bund.bfr.knime.fsklab.preferences.PreferenceInitializer;
+import de.bund.bfr.knime.fsklab.r.server.RConnectionFactory;
+
+public class RprofileManager {
+
+  private static int rHandlerCount;
+  private static ReentrantLock lock = new ReentrantLock();
+  public final static String BFR_R_PLUGIN_NAME= "de.bund.bfr.binary.r.win32.x86_64_3.4.4";
+  
+  public static void subscribe() throws IOException {
+    try {
+      lock.lock();
+      if (rHandlerCount++ == 0 && PreferenceInitializer.getRPath().contains(BFR_R_PLUGIN_NAME)) {
+        RConnectionFactory.createFskLibrary();
+        RConnectionFactory.backupProfile();
+        RConnectionFactory.configureProfile();
+      }
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  public synchronized static void unSubscribe() throws IOException {
+    try {
+      lock.lock();
+      if (--rHandlerCount == 0 && PreferenceInitializer.getRPath().contains(BFR_R_PLUGIN_NAME)) {
+        RConnectionFactory.restoreProfile();
+      }
+    } finally {
+      lock.unlock();
+    }
+
+  }
+
+  
+}
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/RScriptHandler.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/RScriptHandler.java
@@ -5,12 +5,14 @@ import de.bund.bfr.knime.fsklab.nodes.plot.Ggplot2Plotter;
 import de.bund.bfr.knime.fsklab.r.client.IRController.RException;
 import de.bund.bfr.knime.fsklab.r.client.LibRegistry;
 import de.bund.bfr.knime.fsklab.r.client.RController;
+import de.bund.bfr.knime.fsklab.r.client.RprofileManager;
 import de.bund.bfr.knime.fsklab.r.client.ScriptExecutor;
 import de.bund.bfr.knime.fsklab.v2_0.FskPortObject;
 import de.bund.bfr.knime.fsklab.v2_0.FskSimulation;
 import de.bund.bfr.knime.fsklab.v2_0.runner.RunnerNodeInternalSettings;
 import de.bund.bfr.knime.fsklab.v2_0.runner.RunnerNodeSettings;
 import de.bund.bfr.metadata.swagger.Parameter;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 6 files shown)


schuelet:
@ahmadswaid Hi Ahmad, I am still testing but one thing I noticed is that an existing Rprofile that already has a reference to .fsk will remain that reference after the Runner is finished. Although that is expected, users that have this .Rprofile from previous FSK-Lab versions will continue to experience problems. Is there a way to make absolutely sure that after the last Runner is done, this reference is removed no matter what?

schuelet:
@ahmadswaid Two more things (first is minor and possibly ignorable):
- If I change from user R back to FSK-Lab bundle R, I need to restart KNIME, otherwise FSK-Lab tries to access the previously build packages from another R and the Runner fails.
- If the Runner fails, the RProfile will not be reset.

schuelet:
@ahmadswaid Lastly, I have a question about Linux: We want special behaviour here, because there, the bundle is not used but all the installed libraries still are in the .fsk folder and we don't want to lose access to them (some libraries required special commands to get working). Is this considered in this PR?

mfilter:
re the note below:
I wonder, if there is a way how we could ask the user if he agrees to delete the reference to .fsk in his existing .Rprofile? Maybe the user wants to use it even outside of FSK-Lab and then we would "destroy" his setting with each FSK-Lab run.
> 
>
> @ahmadswaid Hi Ahmad, I am still testing but one thing I noticed is that an existing Rprofile that already has a reference to .fsk will retain that reference after the Runner is finished. Although that is expected (we only restore the previous state of Rprofile), users that have this .Rprofile from previous FSK-Lab versions will continue to experience problems. Is there a way to make absolutely sure that after the last Runner is done, this reference is removed no matter what?

ahmadswaid:
@schuelet:
The current PR is taking care of configuring the RProfile for Linux, I tested that on my KNIME VM but I don't know if you have other settings for R on BfR KNIME Sever.
It would be very helpful, if you are able to test this PR there.

@mfilter the current implementation change the the RProfile setting every time we trigger the runner. we could add a global FSK setting next to the R path as check box which says if the user wants to remove the reference to the .fsk with every run.
I can implement that if you agree on the Idea.

mfilter:
Yes, I think this is a good solution.
@schuelet : should we remove the option to provide a path to Python in the FSK settings at this occasion? If I remember right, this is not used anymore.

schuelet:
> Yes, I think this is a good solution.
> @schuelet : should we remove the option to provide a path to Python in the FSK settings at this occasion? If I remember right, this is not used anymore.

@mfilter Yes, good call. 
@ahmadswaid Could you remove the python preference field in the PreferencePage.java, please? Thanks very much.

schuelet:
@ahmadswaid I tested on my Debian VM, and when KNIME is running as sudo, I get the error message:
image

The KNIME Server seems to be running as sudo as well, even though I started the app and executor as user "knime", so I am not sure if my configuration is wrong. I'll try to update the BFR toyserver to double check.

schuelet:
@ahmadswaid On the toyserver it worked. I am not sure, where the error on my machine comes from, but as long as it works on the server, I am fine with it.

ahmadswaid:
the new setting page:
image

schuelet:
@ahmadswaid It looks like it is working but before I approve I will ask @llavall if he could install this on the knimeintern Server so we can test a little bit easier.

ahmadswaid:
> @ahmadswaid It looks like it is working but before I approve I will ask @llavall if he could install this on the knimeintern Server so we can test a little bit easier.

Good Idea, Thank you

schuelet:
@ahmadswaid This error also occurs on the knimeintern server, so we need to investigate.

> @ahmadswaid I tested on my Debian VM, and when KNIME is running as sudo, I get the error message:
> image
> 
> The KNIME Server seems to be running as sudo as well, even though I started the app and executor as user "knime", so I am not sure if my configuration is wrong. I'll try to update the BFR toyserver to double check.

schuelet:
@ahmadswaid Sorry but I have to ask again, does FSK-Lab on Linux have access to the .fsk folder? Because that was what I was talking about last time, we need to have a special case for Linux where the .fsk folder is used even if the user selects their own R.

de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/RprofileManager.java
```diff
@@ -0,0 +1,41 @@
...
+public class RprofileManager {
+
+  private static int rHandlerCount;
+  private static ReentrantLock lock = new ReentrantLock();
+  public final static String BFR_R_PLUGIN_NAME= "de.bund.bfr.binary.r.win32.x86_64_3.4.4";
+  
+  public static void subscribe() throws IOException {
+    try {
+      lock.lock();
+      if (rHandlerCount++ == 0 && (PreferenceInitializer.getRPath().contains(BFR_R_PLUGIN_NAME) || Platform.isLinux())) {
```
ahmadswaid, line 18: @schuelet:
Linux use case is covered in the if

schuelet:
@ahmadswaid Successfully tested with your latest commit. Very good, thanks.

llavall:
> 
> 
> @ahmadswaid I tested on my Debian VM, and when KNIME is running as sudo, I get the error message:
> image
> 
> The KNIME Server seems to be running as sudo as well, even though I started the app and executor as user "knime", so I am not sure if my configuration is wrong. I'll try to update the BFR toyserver to double check.

@schuelet  there is the one main rule of KNIME not to run the KNIME Server as sudo. I am wondering why you say the server is run with sudo rights?

schuelet:
@ahmadswaid I was wrong, it doesn't work. I stupidly compiled the wrong branch. Now after I installed the correct one, the error from above doesn't appear, but the Runner is stuck. I thought it might install some libraries, but that doesn't appear to be the case.

schuelet:
@ahmadswaid update: a model without any library requirements can be executed.

ahmadswaid:
@schuelet: in addition to the list in the description above:
1- .fsk folder is used as a default library path in Linux **with the other pre defined paths** and that mean that during the fsk runner execution any library will  be installed to the .fsk/library folder and RProfile will be restored after finished the run same as the case of windows with BfR R bundle installed.
2- if the user decided to change the R setting to use his own R binary rather than BfR R bundle or the other way around then the LibRegistry instance will be refreshed.

schuelet:
@ahmadswaid would we need to apply all the subscribe() methods to the deprecated RScriptHandler as well in order to avoid confilicts?

ahmadswaid:
> @ahmadswaid would we need to apply the subscribe()/unsubscribe() methods to the deprecated RScriptHandler as well in order to avoid confilicts?

@schuelet: I think so, do you have any workflow with deprecated nodes.

schuelet:
@ahmadswaid here is one: knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/final_test_workflows/Test_deprecatedNodes_v1.9_with%20JoinerWF_ICPMF11_MF_V2

I will keep testing and report when something is coming up.

schuelet:
@ahmadswaid I think we are nearly done, it works on Linux Server and Windows. One last issue I can find: Switching the R installation on Windows requires a KNIME restart, otherwise the Runner complains about an outdated workspace: 
image

schuelet (approved):

PR #913, opened 2021-05-20
workaround for old models

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/334
the model have a model metadata type `GenericModel`  and the expected value is `genericModel` 
I don't know how this model is generated but it seems to be an old model.
I added a workaround like in ConversionUtils:ModelClass enum Line 66


schuelet (approved):
Thanks for the quick fix!

PR #915, opened 2021-05-22
fix https://github.com/RakipInitiative/ModelRepository/issues/335

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/335


schuelet:
@ahmadswaid Sorry it takes so long, but I need more time testing. 
I get an issue when loading a model and just execute the editor without opening it. It seems, the output parameters are included in the simulation.

ahmadswaid:
> @ahmadswaid Sorry it takes so long, but I need more time testing.
> I get an issue when loading a model and just execute the editor without opening it. It seems, the output parameters are included in the simulation.

@schuelet good catch.

schuelet:
@ahmadswaid Found another bug: If there is another simulation present, the second one has the value of the defaultSimulation.

ahmadswaid:
> @ahmadswaid Found another bug: If there is another simulation present, the second one has the values of the defaultSimulation.

@schuelet sorry, everything should be correct by now

schuelet (approved):

PR #920, opened 2021-06-05
Change general information -> reference -> Publication Year datatype to string instead of date

ahmadswaid:


schuelet:
@ahmadswaid I have some questions: Doe changing the portObject affect already saved workflows? Are settings stored in existing Editor nodes still able to load? I will test it anyway, but what is your expectation?

ahmadswaid:
@schuelet saved editor settings is expected to be loaded correctly. also deprecated nodes is expected to be migrated and load correctly.

schuelet:
@ahmadswaid I used the master branch to create a model with reference date 2021, 06, 06. It was saved as 202166 (weird but whatever). Loading the model (Reader) in this branch cuts off the "66" and displays only the year 2021. Maybe the master branch saves the reference date in a wrong format, so this might not be the fault of your changes.
Screenshot 2021-06-07 112820

schuelet:
Here is a workflow so you can quickly check (created with master branch). 
[editor_test.zip](https://github.com/SiLeBAT/FSK-Lab/files/6607699/editor_test.zip)

ahmadswaid:
@schuelet please have another review.

schuelet:
@ahmadswaid The reader doesn't seem to load the full date value, still only the year shows up.

image

Also, the editor with saved settings only loads the year now.

ahmadswaid:
> @ahmadswaid The reader doesn't seem to load the full date value, still only the year shows up.
> 
> image
> 
> Also, the editor with saved settings only loads the year now.

yes, that's correct. it's only the year is allowed to be enterd in this field,
and sorry I thought the lable was changed with the step first of this task

schuelet (approved):
@ahmadswaid I see, then everything is fine. Thx.

PR #921, opened 2021-06-07
executiontime & uploadDate are downloaded in bulk

schuelet:
@ahmadswaid I changed the frontend to fetch the executionTime/uploadDate in one step each. The endpoints were deployed last weekend. 
The loading time went down from 50s to 25s, so a noticable improvement. 25 seconds for a website to load is still too much, but it is a start.


ahmadswaid (approved):
works fine!

PR #926, opened 2021-06-20
FSK DB VIEW: select all feature

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/340
https://github.com/RakipInitiative/ModelRepository/issues/343

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/fskdbview/fskdbview.js
+++ b/de.bund.bfr.knime.fsklab.nodes/js-src/de/bund/bfr/knime/fsklab/v2.0/fskdbview/fskdbview.js
@@ -331,11 +331,27 @@ fskdbview = function () {
                         });
                     }
                 },
+                selectAllRow       : ( O, rowIndex, rowData ) => {
+                    this.checked = true;
+                    $(this).closest("tr").css("background-color", "#e1e3e8");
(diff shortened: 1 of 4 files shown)


schuelet:
@ahmadswaid a few issues:
- filtering for a subset of models -> select all -> all models, including those not shown after filtering, are selected(true)
- option to select max number of models is still between 2 and 4 (I can live with that though)
- performance is very slow, FSK-Web endpoint (>400 models) crashes when using select-all checkbox

The use-case is to search  for gropin models in landingpage/FSK-Web/ and select them all, without the other models

ahmadswaid:
> @ahmadswaid a few issues:
> 
>     * filtering for a subset of models -> select all -> all models, including those not shown after filtering, are selected(true)
> 
>     * option to select max number of models is still between 2 and 4 (I can live with that though)
> 
>     * performance is very slow, FSK-Web endpoint (>400 models) crashes when using select-all checkbox
> 
> 
> The use-case is to search for gropin models in landingpage/FSK-Web/ and select them all, without the other models
regarding the first point, I was preparing a solution in another PR, I'll cancel it and add it to the current PR.
second point, these setting is only used in the joiner component.

ahmadswaid:
@schuelet: fixed performance issue and selection of filtered rows only if filter is applied.
please have another review.

schuelet (approved):
great

PR #928, opened 2021-06-21
Force terminate unused controllers

schuelet:
- There was a library check that tested each library if it is valid and that blew up the Rserve workspace everytime a model was running. With our new Rprofile implementation, that is now completely superfluous.
- More important, a new LibRegistry instance (and therefore a new RController) was created every time a Runner was executed. That meant, the old instance was not shut down properly. It is now ensured, that only one LibRegistry is open while needed.
- Also, each Rserve instance is now almost immeadiatly closed after executon (100ms). The 60000ms might have been useful for KNIME's R integration but we don't need that.

This massively improves memory management without changing too much of the code.

### DIFF
--- a/de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/LibRegistry.java
+++ b/de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/LibRegistry.java
@@ -136,7 +136,8 @@ public class LibRegistry {
 
   public synchronized static LibRegistry instance() throws IOException, RException {
     if (instance == null || PreferenceInitializer.refresh) {
-      instance = new LibRegistry();
+      if(instance == null)
+    	instance = new LibRegistry();
       PreferenceInitializer.refresh = false;
     }
     return instance;
--- a/de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/LibRegistry.java
+++ b/de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/LibRegistry.java
@@ -109,17 +109,7 @@ public class LibRegistry {
       String[] pkgArray = installPath.toFile().list();
       installedLibs = Arrays.stream(pkgArray).collect(Collectors.toSet());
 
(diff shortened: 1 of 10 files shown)


de.bund.bfr.knime.fsklab.r/src/de/bund/bfr/knime/fsklab/r/client/LibRegistry.java
```diff
@@ -135,7 +126,7 @@ private LibRegistry() throws IOException, RException {
   }
 
   public synchronized static LibRegistry instance() throws IOException, RException {
-    if (instance == null || PreferenceInitializer.refresh) {
```
ahmadswaid, line 138: @schuelet: removing this check will prevent the recreation of the LibRegistry instance, which is needed when the user changes the global FSK setting(this is part of RProfile managment)
why it's necessary to remove the condition?

ahmadswaid: I see you description now, but have you tested it with changing the global setting?

schuelet: > @schuelet: removing this check will prevent the recreation of the LibRegistry instance, which is needed when the user changes the global FSK setting(this is part of RProfile managment)
> why it's necessary to remove the condition?

@ahmadswaid Each time an RController is created, PreferenceInitializer.refresh is set to true. See RController.initR() ->
`m_rProps =RBinUtil.retrieveRProperties();` . Therefore, each time a new LibRegistry instance will be created.

I do see your point, so I will try to find a workaround.

schuelet:
- @ahmadswaid I resolved the issues. The PreferenceInitializer.refresh is now set to true in the PreferencePage. When R is changed in the preferences, the LibRegistry controller is closed and a new LibRegistry instance is created (new Runner nodes wait until that is completed). 
- Also, deprecated Writer nodes now close the Scripthandler properly. I hope I thought of everything.
- For some reason I had to increase the time to close the controllers to ~10 seconds, I don't know the reason, but 100 ms was not enough and locked the nodes permanently.

ahmadswaid (approved):
perfect, tested also from my side, every thing worked fine

PR #930, opened 2021-06-24
clean generated resources on reset

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/216


schuelet:
@ahmadswaid this would only work for simple models, not combined ones, right?

ahmadswaid:
> @ahmadswaid this would only work for simple models, not combined ones, right?

yes

ahmadswaid:
@schuelet now the PR covers also the Joined models.

schuelet (approved):

PR #931, opened 2021-06-25
check if validation errors exists before manipulating

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/346


schuelet:
@ahmadswaid Can you explain what you did here? It doesn't look like much has changed, why did only the models from that workflow throw this error?

ahmadswaid:
> @ahmadswaid Can you explain what you did here? It doesn't look like much has changed, why did only the models from that workflow throw this error?

it seemd that models in this workflow have not issue with the metadata so validation map was null so that it was giving an error when saving.
do you see the changes in the app_1.0.0.js It could be that you need to to trigger a maven build on the de.bund.bfr.knime.js project to have the bundle generated again.

schuelet:
@ahmadswaid Oh, no, it did work fine, I was just wondering what the issue was. Thanks for clarifying that.

schuelet (approved):

PR #932, opened 2021-06-28
Improve json handling

schuelet:
When a model creates parameter data during execution, it used to be stored in a java object to be serialized later into a parameters.json file. This was memory inefficient, so the parameter data for a combined model is now streamed into a file directly without holding it in an object.
This will be done for single models as well, but in a later PR.

Also: 
- JsonHandler.applyJoinRelation was improved to only open a parameters.json file once instead of multiple times.
- Ggplot2Plotter had a weird size limit of the plot (4x4), which lead to cropped images. Now there is no size limit.
- Ggplot2Plotter also was bugging out if "import(ggplot2)" was written somewhere in the model script. Now it seems to be working (not related to https://github.com/RakipInitiative/ModelRepository/issues/330)
- LibRegistry had a nullPointerException bug when a user switches the R installation before running a model with the bundled R.
- Sometimes there was a an issue with R packages not getting installed which lead to LibRegistry killing the Rserve Controller and preventing subsequent Controllers from installing packages. This was changed in RScripthandler.installLibs where an unsuccessful installation will lead to creating a new LibRegistry with a new controller.

The PR can best be tested with knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/Joining_TrichinellaModels

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
@@ -8,6 +8,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -57,6 +58,7 @@ public abstract class JsonHandler {
     }
   }
 
+  
   // Method loads parameter data from previous models into workspace of current model
   // according to the list of Joiner-Relations
   public void applyJoinRelation(FskPortObject fskObj, List<JoinRelationAdvanced> joinRelationList,
@@ -66,28 +68,19 @@ public abstract class JsonHandler {
     // a json file twice
     if (joinRelationList != null) {
       // get all relevant parameters -> assign to Map with key=jsonPath
-      Map<String, List<JoinerObject>> sourceTargetPathMap = new HashMap();
+      Map<String, List<JoinerObject>> sourceTargetPathMap = new HashMap<String, List<JoinerObject>>();
       for (JoinRelationAdvanced joinRelation : joinRelationList) {
-        String targetParameter = joinRelation.getTargetParam();
-        FskPortObject sourceModel = joinRelation.getModel();
-        String sourceParameter = joinRelation.getSourceParam();
-
-        for (Parameter param : SwaggerUtil.getParameter(fskObj.modelMetadata)) {
-          if (targetParameter.equals(param.getId() + suffix)) {
-            String resourcePath = sourceModel.getGeneratedResourcesDirectory().get()
-                .getAbsolutePath().replaceAll("\\\\", "/") + "/";
-            String jsonPath = resourcePath + JSON_FILE_NAME;
-            if(!sourceTargetPathMap.containsKey(jsonPath)) {
-              sourceTargetPathMap.put(jsonPath, new ArrayList<JoinerObject>());
+        if(SwaggerUtil.getParameter(fskObj.modelMetadata) != null) {
+          for (Parameter param : SwaggerUtil.getParameter(fskObj.modelMetadata)) {
+            if (joinRelation.getTargetParam().equals(param.getId() + suffix)) {
+              addEntryToRelationMap(sourceTargetPathMap,joinRelation, param );
+    
             }
-
-            sourceTargetPathMap.get(jsonPath).add(new JoinerObject(sourceParameter,param,joinRelation, resourcePath));
-
           }
         }
       }// for joinRelation
 
-
+//TODO: check what happens if sourceTargetPathMap is empty
       // work through Map, each entry is a parameters.json file, load data into workspace
       for (Map.Entry<String, List<JoinerObject>> entry : sourceTargetPathMap.entrySet()) {
         //ParameterData parameterData = MAPPER.readValue(new File(jsonPath), ParameterData.class);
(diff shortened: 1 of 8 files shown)


ahmadswaid (commented):
@Thomas, found an issue when trying to run the joined model using the Risk characterisation model as the last one.

`ERROR FSK Runner           0:15       Execute failed: WARNING: OUTPUT parameter  df_summary of model 02592227-3f31-4f75-ab4e-97ef91cd931c not available in workspace after execution. Make sure all output variables are global in the model script.`

I don't know if the issue is related to the models in this worklflow because I don't see the issue when running some other joiner workflow.

image

schuelet:
@ahmadswaid Thanks for having a look at this unwieldy PR. I can't reproduce the error, which R version do you use? The modules don't run on 3.4.4, so I used R 4.0.4.
Does the Meat Production model run (with JSON checkbox enabled) through?

ahmadswaid:
> modules don't run on 3.4.4, so I used R 4.0.4.

I used R version 4.1.0 on both MacOS and Windows. and just run successfully the Meat Production model with check box enabled.

schuelet:
@miguelalba Can you run a test on your machine with this branch and the above workflow? When you find the time, thanks.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
```diff
@@ -52,11 +57,16 @@ protected JsonHandler(ScriptHandler scriptHandler, ExecutionContext exec) {
...
+  // Method loads parameter data from previous models into workspace of current model
+  // according to the list of Joiner-Relations
   public void applyJoinRelation(FskPortObject fskObj, List<JoinRelationAdvanced> joinRelationList,
       String suffix) throws Exception {
 
+    // figure out which parameters from JoinRelations belong to which model to avoid loading
+    // a json file twice
     if (joinRelationList != null) {
+      // get all relevant parameters -> assign to Map with key=jsonPath
+      Map<String, List<JoinerObject>> sourceTargetPathMap = new HashMap();
```
miguelalba, line 69: Type safety. Missing diamond operator.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
```diff
@@ -52,11 +57,16 @@ protected JsonHandler(ScriptHandler scriptHandler, ExecutionContext exec) {
...
       for (JoinRelationAdvanced joinRelation : joinRelationList) {
         String targetParameter = joinRelation.getTargetParam();
```
miguelalba, line 71: Personal preference. If getters do not have any logic I would use them straight instead of assigning them to local variables to keep the stack clean.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
```diff
@@ -67,22 +77,53 @@ public void applyJoinRelation(FskPortObject fskObj, List<JoinRelationAdvanced> j
             String resourcePath = sourceModel.getGeneratedResourcesDirectory().get()
                 .getAbsolutePath().replaceAll("\\\\", "/") + "/";
             String jsonPath = resourcePath + JSON_FILE_NAME;
-            loadParametersIntoWorkspace(jsonPath, sourceParameter, param.getId());
```
miguelalba, line 70: I cannot write on 75 since it did not change but there is a NPE there. SwaggerUtil.getParameter may return null and needs to be checked.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
```diff
@@ -67,22 +77,53 @@ public void applyJoinRelation(FskPortObject fskObj, List<JoinRelationAdvanced> j
             String resourcePath = sourceModel.getGeneratedResourcesDirectory().get()
```
miguelalba, line 77: getGeneratedResourcesDirectory() returns an Optional that may be empty and needs to be checked.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
```diff
@@ -67,22 +77,53 @@ public void applyJoinRelation(FskPortObject fskObj, List<JoinRelationAdvanced> j
...
                 .getAbsolutePath().replaceAll("\\\\", "/") + "/";
             String jsonPath = resourcePath + JSON_FILE_NAME;
-            loadParametersIntoWorkspace(jsonPath, sourceParameter, param.getId());
-
-            // if target parameter is of type FILE, add path to generatedResources to sourceParam
-            // This should be safe since source and target parameter must have the same type (if
-            // File)
-            if (param.getDataType().equals(Parameter.DataTypeEnum.FILE)) {
-              addPathToFileParameter(param.getId(), resourcePath);
+            if(!sourceTargetPathMap.containsKey(jsonPath)) {
```
miguelalba, line 80: This is more a matter of style. If sourceTargetPath does not contain jsonPath a new List of JoinerObject is added. In any case a new JoinerObject is added to the list.

Here I would create the joiner object first because it'll be used in both cases.
```java
JoinerObject joinerObject = new JoinerObject(sourceParameter,param,joinRelation, resourcePath)
```
Then I'd just add joinerObject or initialize the list with it according to the if:
```java
if (!sourceTargetPathMap.containsKey(jsonPath)) {
    sourceTargetPathMap.put(jsonPath, Arrays.asList(joinerObject));
} else {
    sourceTargetPathMap.get(jsonPath).add(joinerObject);
}
```

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/JsonHandler.java
```diff
@@ -67,22 +77,53 @@ public void applyJoinRelation(FskPortObject fskObj, List<JoinRelationAdvanced> j
...
+
           }
         }
+      }// for joinRelation
+
+
+      // work through Map, each entry is a parameters.json file, load data into workspace
+      for (Map.Entry<String, List<JoinerObject>> entry : sourceTargetPathMap.entrySet()) {
+        //ParameterData parameterData = MAPPER.readValue(new File(jsonPath), ParameterData.class);
+        ParameterData parameterData = MAPPER.readValue(new FileInputStream(entry.getKey()), ParameterData.class);
```
miguelalba, line 94: FileInputStream is not closed. This could be either opened in a try-with-resources statement or just replaced with File.
I have a question about jsonPath. If they are only used here, could they be replaced with File objects? It would save you handling the file paths and you could just iterated through file keys.

miguelalba (changes requested):
Changes are needed. I still need to run workflow and give feedback.

PR #933, opened 2021-07-06
Stream json parameters

schuelet:
When a model creates parameter data during execution, it used to be stored in a java object to be serialized later into a parameters.json file. This was memory inefficient, so the parameter data for a combined model is now streamed into a file directly without holding it in an object.
This will be done for single models as well, but in a later PR.

Also: 
- JsonHandler.applyJoinRelation was improved to only open a parameters.json file once instead of multiple times.
- Ggplot2Plotter had a weird size limit of the plot (4x4), which lead to cropped images. Now there is no size limit.
- Ggplot2Plotter also was bugging out if "import(ggplot2)" was written somewhere in the model script. Now it seems to be working (not related to https://github.com/RakipInitiative/ModelRepository/issues/330)
- LibRegistry had a nullPointerException bug when a user switches the R installation before running a model with the bundled R.
- Sometimes there was a an issue with R packages not getting installed which lead to LibRegistry killing the Rserve Controller and preventing subsequent Controllers from installing packages. This was changed in RScripthandler.installLibs where an unsuccessful installation will lead to creating a new LibRegistry with a new controller.

The PR can best be tested with knime://external-KNIME-Server/FSKX_Testing/FSK-Lab_2.0/broken_workflows/Joining_TrichinellaModels


schuelet:
This is a new branch, the other was scrapped. This is a short term implementation of streaming json data to a file, but it still loads the data of single parameters into a java object. When I have the time, the entire data will be streamed to files, to prevent heap space errors, but for now this will hopefully work for most use-cases.

ahmadswaid (approved):
@schuelet: great work, I will approve it but I still have a problem in running the Risk characterisation model in the referenced workflow, I know it's not related to this PR but we should investigate this model, maybe if @miguelalba can confirm my finding we could spend more time on it.
tested also the simple, multiple, cascade joining and that worked fine in addition to the joiner component with local and remote repository.

PR #946, opened 2021-08-13
Readme in detailsview in fskdb view

ahmadswaid:
@schuelet  please note that I assumed that you will create an endpoint called `readme`
please let me know to change it if it dosen't work for you


schuelet:
@ahmadswaid thanks, it works very well. Is there an easy way to wrap the lines, so they don't extend horizontally beyond the limits of the modal view? The new backend will be deployed by tomorrow. 

image

ahmadswaid:
@schuelet I wonder how I can reproduce your test, I used the visualzation script endpoint instead of readme endpoint for the sake of testing and the readme panel showed the line breaks correctly.
image

schuelet:
@ahmadswaid The endpoint /readme is online. But regarding the script, those are also not wrapping, you can see it in some model scripts. But those are ok, since we want the line indices anyway. In the Readme, we could do without them and force a text-wrap class to the span element. When I do that, currently it will look like this:

image

Can we somehow get rid of the line numbers for the Readme Tab only?

schuelet:
@ahmadswaid If that is too much headache, I can live without it right now. No big deal.

ahmadswaid:
> @ahmadswaid If that is too much headache, I can live without it right now. No big deal.
@schuelet please have another check, here it's how it looks on my device:
image

schuelet (approved):
great, can't thank you enough.

PR #1041, opened 2022-06-02
Runner Offline Support

ahmadswaid:
postpone the Internet connectivity check, untill finding that the package is not available locally.


schuelet:
Hi @ahmadswaid , thanks a lot for the quick response. One question, why did you need to alter the classpath in two projects? Were these leftovers that needed to be changed for Java 11? Can I ignore the changes when merging with target 4.0 ?

ahmadswaid:
> Hi @ahmadswaid , thanks a lot for the quick response. One question, why did you need to alter the classpath in two projects? Were these leftovers that needed to be changed for Java 11? Can I ignore the changes when merging with target 4.0 ?

yes, it's something had to be done before for eclipse to be able use java11 features. so as you said please ignore it for merging with Knime 4.0

PR #1044, opened 2022-06-10
fixing https://github.com/RakipInitiative/ModelRepository/issues/409

ahmadswaid:


schuelet:
@ahmadswaid It works when I create a new simulation and give it a name. But if the user neglects to name the new simulation, the Runner fails again. I can live with that, but is there a quick way to generate a name in case there is none (or completely discard the new simulation in that case)?

ahmadswaid:
> @ahmadswaid It works when I create a new simulation and give it a name. But if the user neglects to name the new simulation, the Runner fails again. I can live with that, but is there a quick way to generate a name in case there is none (or completely discard the new simulation in that case)?

@schuelet pushed a new commit to cover this usecase please have another check!

ahmadswaid:
> > @ahmadswaid It works when I create a new simulation and give it a name. But if the user neglects to name the new simulation, the Runner fails again. I can live with that, but is there a quick way to generate a name in case there is none (or completely discard the new simulation in that case)?
> 
> @schuelet pushed a new commit to cover this usecase please have another check!

only one note, please rebuild the javascript de.bund.bfr.knime.js or the the wole FSKLab modules to have the changes in place.

PR #1056, opened 2022-07-13
fix a Joiner Bug: Nested Joining shows undefined parameters

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/412

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/joiner/JoinerNodeModel.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/joiner/JoinerNodeModel.java
@@ -896,8 +896,7 @@ public final class JoinerNodeModel extends
             && unModifiedParamsNames.containsKey(firstModelName)) {
           resetParameterIdToOriginal(
               SwaggerUtil.getParameter(fskID_to_fskObject.get(firstModelName).modelMetadata),
-              ++suffixIndex, true);
-          --suffixIndex;
+              suffixIndex, true);
         } else {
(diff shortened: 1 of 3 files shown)


schuelet:
@ahmadswaid I have found an issue here. In this workflow, joining a combined model with another model leads to an error. I found out so far, that the secondFSKObject of the first combinedObject does not generate a simulation ( fskSimulationSecond = [ ] ). Don't know why though.
[joiningCombinedModelsNew.zip](https://github.com/SiLeBAT/FSK-Lab/files/9403311/joiningCombinedModelsNew.zip)

ahmadswaid:
@schuelet fixed an issue with the generation of the JoinRelationAdvanced object out of the joiner relation as a preparation step before running the combined object.
Please review.

schuelet:
@ahmadswaid this throws an error in one of the joiner test workflows as well:
[Test_Comb_Models_Chain (2).zip](https://github.com/SiLeBAT/FSK-Lab/files/9453309/Test_Comb_Models_Chain.2.zip)

I am not sure if this suffix handling is the play anymore. The code has become so outlandishly convoluted that I am considering to try and find a better way to handle the joining. We also run into trouble explaining the sim.sedml files to others, I am thinking, we could try to find a way to use proper references instead of just parameter id's. Look at my ticket RakipInitiative/ModelRepository#419, which is also quite a challenge since the suffix system is so clunky. Need to discuss with @mfilter

ahmadswaid:
Hi @schuelet 
Thanks for you investigation. Please look into the new improvement.
I agree the code for this case is complex but I think that big part of the complexity is of the use case itself.
I will add code documentation that will simplify the complixity as possible.
but anyway I'm open to any new Idea.

schuelet (approved):

PR #1058, opened 2022-08-09
optimize preference page

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/417


schuelet:
@ahmadswaid This is much better, thanks. There is an issue though, the python environments are not "locked in" unless I click on the "check environment" button. If that click-event is mandatory for the python stuff to work, it would be ok to have a warning text to remind the user to click the button to confirm (though I would expect the button "close & apply" to do that).

I am also not completely happy with the delay when the preference page is opened and then again once the R environment is checked (to test the python environments). It still seems that there are two separate full testing routines, one for R environments and one for the python 2 / 3 envs. Can we not cover all of them in the same turn?

ahmadswaid:
> @ahmadswaid This is much better, thanks. There is an issue though, the python environments are not "locked in" unless I click on the "check environment" button. If that click-event is mandatory for the python stuff to work, it would be ok to have a warning text to remind the user to click the button to confirm (though I would expect the button "close & apply" to do that).
> 
> I am also not completely happy with the delay when the preference page is opened and then again once the R environment is checked (to test the python environments). It still seems that there are two separate full testing routines, one for R environments and one for the python 2 / 3 envs. Can we not cover all of them in the same turn?

@schuelet you're quite correct, actually it's not only a performance issue but also not a correct behaviour to save selected environments only on testing.
Thanks for investigation.

schuelet (approved):

PR #1064, opened 2022-10-12
Fix bug that modal is always on top of component view

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/354


schuelet:
@ahmadswaid: I like your solution because it is simple and won't break any of the business logic. However, by auto-scrolling to the top of the page we run into trouble when we have other JS-Views in the Component (e.g. the Joiner on top and DBView on bottom). Is there a way to catch this case (multiple views)?

ahmadswaid:
@schuelet done, please review it. rebuild of JS module is required!

schuelet:
@ahmadswaid : Only works the first time I click on the details button. After that, the view is all over the place. Each time I move the mouse or click someplace, the vertical scrollbar slider freaks out.

schuelet (approved):
@ahmadswaid All clear. Thanks a lot!

PR #1065, opened 2022-10-19
move all temporary data to a user specified folder

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/413

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes.common/META-INF/MANIFEST.MF
+++ b/de.bund.bfr.knime.fsklab.nodes.common/META-INF/MANIFEST.MF
@@ -14,7 +14,8 @@ Require-Bundle: com.fifesoft.rsyntaxtextarea;bundle-version="2.6.0",
  org.eclipse.core.runtime;bundle-version="3.13.0",
  org.knime.core;bundle-version="[4.0.0,5.0.0)",
  org.knime.python2;bundle-version="[4.0.0,5.0.0)",
- org.w3c.dom.svg;bundle-version="1.1.0"
+ org.w3c.dom.svg;bundle-version="1.1.0",
+ de.bund.bfr.knime.fsklab.preferences;bundle-version="1.0.0"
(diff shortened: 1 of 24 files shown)


schuelet:
@ahmadswaid That is super interesting, to see all the files being generated in a single folder. Especially, since most of those files don't seem to be deleted when the workflow is reset or closed. I presume, some of the garbage collection is done by KNIME and doesn't apply to the user defined folder, but still. There are also an awful lot of workspace.r files generated, I wonder if we should investigate this further. 

I will do some more testing, also in a KNIME server environment, before this can go live, but it looks good so far.

PS: There a strange error when using the DBView in a component (it doesn't actually fail the node):
image

Was that error there before?

ahmadswaid:
> @ahmadswaid That is super interesting, to see all the files being generated in a single folder. Especially, since most of those files don't seem to be deleted when the workflow is reset or closed. I presume, some of the garbage collection is done by KNIME and doesn't apply to the user defined folder, but still. There are also an awful lot of workspace.r files generated, I wonder if we should investigate this further.
> 
> I will do some more testing, also in a KNIME server environment, before this can go live, but it looks good so far.
> 
> PS: There a strange error when using the DBView in a component (it doesn't actually fail the node): image
> 
> Was that error there before?

@schuelet That is fair. I will do some testing also because I know it has huge impact at the end.
regarding the DBView issue, I didn't see it before. I will check it and open a new ticket if needed.

schuelet:
> @schuelet That is fair. I will do some testing also because I know it has huge impact at the end. regarding the DBView issue, I didn't see it before. I will check it and open a new ticket if needed.
@ahmadswaid reproducible by using one of the joiner example workflows, e.g. (knime://knime.mountpoint/RAKIP-Web/joining_examples/JoinerComponent_Single_Hit_Risk_Population) . It has been part of FSK-Lab for an unknown period of time, but when using a standalone KNIME, it only shows up as a warning, not an error.
image

schuelet (approved):

PR #1066, opened 2022-10-22
Makes files associated to FSKX model visible in port object

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/385

### DIFF
--- a/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/environment/AddedFilesEnvironmentManager.java
+++ b/de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/environment/AddedFilesEnvironmentManager.java
@@ -23,7 +23,7 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
 public class AddedFilesEnvironmentManager implements EnvironmentManager {
 
   private EnvironmentManager manager;
-  private String[] files;
+  private String[] entries;
   private List<Path> markedForRemoval = new ArrayList<Path>();
   
   public AddedFilesEnvironmentManager() {
@@ -35,9 +35,9 @@ public class AddedFilesEnvironmentManager implements EnvironmentManager {
   public AddedFilesEnvironmentManager(String[] files) {
     this(new DefaultEnvironmentManager(), files);
   }
-  public AddedFilesEnvironmentManager(EnvironmentManager manager, String[] files) {
+  public AddedFilesEnvironmentManager(EnvironmentManager manager, String[] entries) {
     this.manager = manager;
-    this.files = files;
+    this.entries = entries;
     
   }
   
@@ -52,7 +52,7 @@ public class AddedFilesEnvironmentManager implements EnvironmentManager {
   public Optional<Path> getEnvironment() {
     Optional<Path> environment = manager.getEnvironment();
     
-    if (files == null || files.length == 0)
+    if (entries == null || entries.length == 0)
       return environment;
  
     return copyFilesToEnvironment(environment);
@@ -62,7 +62,7 @@ public class AddedFilesEnvironmentManager implements EnvironmentManager {
     
     List<Path> filePaths = new ArrayList<>();
     try {
-      for (String filePath : files) {
+      for (String filePath : entries) {
         filePaths.add(FileUtil.resolveToPath(FileUtil.toURL(filePath)));
       }
 
@@ -114,7 +114,7 @@ public class AddedFilesEnvironmentManager implements EnvironmentManager {
   
   @Override
   public String[] getEntries() {
-    return files;
+    return entries;
   }
     
 }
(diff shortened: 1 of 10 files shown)


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/CombinedFskPortObject.java
```diff
@@ -702,10 +705,33 @@ public JComponent[] getViews() {
...
+    if(fskObject instanceof CombinedFskPortObject) {
+      fillResoucesList(((CombinedFskPortObject)fskObject).getFirstFskPortObject(), resourcesList);
+      fillResoucesList(((CombinedFskPortObject)fskObject).getSecondFskPortObject(), resourcesList);
+    }else {
+      if (fskObject.getEnvironmentManager().isPresent()) {
+        String modelName = SwaggerUtil.getModelName(fskObject.modelMetadata);
+        String padding = String.join("", Collections.nCopies(25, "-"));
+        resourcesList.add(padding+modelName+padding);
+        
+        List<String> elements = Arrays.asList(((ArchivedEnvironmentManager) fskObject.getEnvironmentManager().get()).getEntries());
```
schuelet: @ahmadswaid Seems to work great, but here (and in FSKPortObject.java) you cast to ArchivedEnvrionmentManager, which throws an error when a model is created using the editor. I suggest to either add cases for the other Manager Classes (FilesEnvironment,AddedFilesEnvironment and ExistingEnvironment) or add a getEntries Method to the Interface (probably more elegant but you decide).

schuelet (changes requested):

ahmadswaid:
@schuelet  Thanks for the hint, implemented! please have another review

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/environment/ExistingEnvironmentManager.java
```diff
@@ -62,4 +63,16 @@ public Optional<Path> getEnvironment() {
   public void deleteEnvironment(Path path) {
     // The path existed previously and should not be removed.
   }
+
+  @Override
+  public String[] getEntries() {
```
schuelet, line 69: @ahmadswaid: The Mapper complains about the missing property "entries": 
image

schuelet (changes requested):

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/environment/AddedFilesEnvironmentManager.java
```diff
@@ -115,5 +111,10 @@ public void deleteEnvironment(Path path) {
     clearAddedFiles();
     manager.deleteEnvironment(path);
   }
+  
+  @Override
+  public String[] getEntries() {
+    return entries;
```
schuelet, line 117: @ahmadswaid: this method returns only the added files. The other environmentManager classes return all files associated with getEnvironment(). Fixing this isn't easy. If you change the method to return added files AND the files from the "sub-manager", then you need to change the handling of added files in FSKEditorJSNodeDialog as well. I was unable to do that, somehow on clicking "ok" in the dialog, all the entries were duplicated. I pushed my changes to a separate branch, maybe you can figure out what to do here. [see here](https://github.com/SiLeBAT/FSK-Lab/tree/addedFiles_problem)


1st Editor: add directory
image

2nd Editor, Add File "hello.py"
image

On reopening the Dialog:
image

schuelet (changes requested):
*edit: comment deleted.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/nodes/environment/AddedFilesEnvironmentManager.java
ahmadswaid: Thanks @schuelet, I noticed this, I'm working on it, because of that I didn't ask for review :)
Actually it seems to be an old problem.

ahmadswaid: @schuelet this should be ready, please have the review. thxs

schuelet: @ahmadswaid the idea of the AddedFilesManager was to allow to add files to an existing model. Miguels Environment implementation didn't allow for that use case, so I tried to build an environment that carries its "parent environment" and in addition tracks the files that were added by the "add files" button. 
As of now, the files added by using this "add file" button overwrite the existing files, that's what I meant in an earlier comment ("getEntries "returns only the added files", not all files from existing environments").
image

image


 I was unable to fix it (see images and branch above), and unfortunately the issue remains in the latest commit.

ahmadswaid:
@schuelet no duplication/ adding file to inherited list of resources without clearing it is possible.
I tried my best to test the cases. please have another look.

schuelet:
> @schuelet no duplication/ adding file to inherited list of resources without clearing it is possible. I tried my best to test the cases. please have another look.

@ahmadswaid in the latest commit, the editor seems to clear everything from the "Resources List" tab. Chaining two editors also removes all entries, regardless whether I added any files or not. In fact, connecting any node, e.g. a Simulation Configurator to a Reader removes all entries.
Maybe it is best to put this ticket on ice and focus on more important ones. Too bad, it seemed we almost had a working solution, if it weren't for the "add file" use-case.

schuelet (approved):

PR #1067, opened 2022-10-28
Jgit versioning

schuelet:
@ahmadswaid Tycho version 2.7.3. On maven clean verify, the build throws the error discussed here: https://forum.knime.com/t/preview-knime-ap-4-4-java-11-eclipse-upgrade/32587


ahmadswaid:
@schuelet work is still in progress. but now I migrated the pmm.common plugin to tycho 2.7.3
Next step is migrating the de.bund.bfr.knime.pmm.nodes which has the same nature of building issues but now these are easier to be solved as I know the solution (use imported packages instead of required plug-ins when possible) as this will not import more packages than what is needed.

schuelet:
@ahmadswaid great! Two questions though: 1. Do you expect any changes in PMM or FPL since they use slightly different dependencies?
2. Can you quickly have a second look at the feature paths? They seem to be build into 2 separate groups: 
image

ahmadswaid:
@schuelet:
1- regarding question one, not too much since the SDK doesn't changes in between for the methods that we use, though there were a little changes done in the code to reflect the new changes in the API like:

https://github.com/SiLeBAT/FSK-Lab/commit/d43b7686b1f2f5d1acc85ddeea5f15789c581745#diff-7f517deae0261ec2df85a0b9e5e2acc24d9c2e3f79a2bba196c5b1b08424bbc2L1037

And here:
https://github.com/SiLeBAT/FSK-Lab/commit/26ab053895bfdd88bd23a757e08d2107f4e18d71#diff-28d8b4b64cb1ff1de59a1379e261964adf66edba531224a09e4d943df1b46bd9L108

2 - regards point two I though, it's something you intended:
https://github.com/SiLeBAT/FSK-Lab/commit/7dc5cb484a93da8ed81a1ec9c20c9967997f2226#diff-554de61172182e16fe3313991b74b6656d3a7ec03e7ffa2b0757ddc717b45fc4L6

any way I will fix it back!

PR #1068, opened 2022-11-07
Editor bug switching model category in js view does not change view

ahmadswaid:
https://github.com/RakipInitiative/ModelRepository/issues/305

With this commit and PR, I remove the "Close & Apply" on changing the model class. 
This way we are sure that the user have only the chance to cancel or save the setting as a new defaults. and witht his we garantuee that the model class is transfered to the dialog.

@mfilter @schuelet if you agree on the suggested solution, I will next test it on the Knime server.


schuelet:
@ahmadswaid : Found two minor bugs:
- On switching the model category, the creation date is reset to tt.mm.jj
- On switching to Dose-Response Model or switching from Dose-Response-Model, the model name is removed:
image

ahmadswaid:
@schuelet solved mentioned issues and the model class selection after conversion.
Please have another review.

schuelet (approved):

PR #1084, opened 2023-01-13
Pmm converter

ahmadswaid:
fixing issue:
https://github.com/RakipInitiative/ModelRepository/issues/428
Thanks @TarGue for support.
@schuelet could you please review this. Thanks


schuelet:
@ahmadswaid I will create a new PR and only include the changes that make sense. The BasePlotter for example should not be changed. 
Also a question: It appears, that the metadata is not converted to FSKX file except the name: Is that in agreement with Matthias?

PR #1110, opened 2023-03-12
Feature request infuse simulation into sim config via flow variable

ahmadswaid:
This way any file with URI scheme used in the begining like : http, https, ftp, file:// will be downloaded 
the simulation can be provided as a json object where parameters and values provided as pair like
{
    "name": "value"
}
if the name doesn't exist in the parameter list of the model then it'll be ignored.


schuelet:
@ahmadswaid Here the workflow, the model is downloaded by the Reader since it is quite large:
 
[infuse_data_into_simulation.zip](https://github.com/SiLeBAT/FSK-Lab/files/10959036/infuse_data_into_simulation.zip)

schuelet:
@ahmadswaid Also, I tried to use this URL from Zenodo which has a parameter which screws up resolving the file-name:
https://zenodo.org/record/7651129/files/covid19za_provincial_cumulative_timeline_confirmed.csv?download=1

PR #1124, opened 2024-02-08
Add python packages to package.json

ahmadswaid:
Extract python distribution version and the packages version used in when creating the models.
The PythonScriptAnalyser is part of fskml repository (https://github.com/SiLeBAT/fskml) but the source code lib is added to the libs folder


de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/editor/FSKEditorJSNodeModel.java
```diff
@@ -524,10 +531,55 @@ protected PortObject[] performExecute(PortObject[] inObjects, ExecutionContext e
...
+        String languageWrittenIn = SwaggerUtil.getLanguageWrittenIn(metadata);
+        String pythonPath = null; // Path to the Python executable in the conda environment
+        String envPath = null; // Path to the conda environment
+        if(PreferencePage.getPython3CondaEnvironmentDirectoryPath() != null)
+          envPath = PreferenceInitializer.getPython3Env();
+        else
+          envPath = PreferenceInitializer.getPython2Env();
+        
+        if(((FskPortObject)inObjects[0]).overridePackages) {
+          pythonPath =envPath+"/bin/python";
```
schuelet, line 544: This path doesn't work for Windows. There it would be more like:
`          pythonPath =Path.of(envPath, "python.exe").toString();
`

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/writer/WriterNodeModel.java
```diff
@@ -525,25 +535,84 @@ private static void writeArchive(File archiveFile, FskPortObject portObject,
...
+      if(portObject.modelMetadata != null && SwaggerUtil.getLanguageWrittenIn(portObject.modelMetadata) != null) {
+        String languageWrittenIn = SwaggerUtil.getLanguageWrittenIn(portObject.modelMetadata);
+        String pythonPath = null; // Path to the Python executable in the conda environment
+        String envPath = null; // Path to the conda environment
+        if(PreferencePage.getPython3CondaEnvironmentDirectoryPath() != null)
+          envPath = PreferenceInitializer.getPython3Env();
+        else
+          envPath = PreferenceInitializer.getPython2Env();
+        
+        pythonPath =envPath+"/bin/python";
```
schuelet, line 549: Same problem as in FSKEditorJSNodemModel.java (line 544), in Windows it doesn't work.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/editor/FSKEditorJSNodeModel.java
```diff
@@ -524,10 +531,55 @@ protected PortObject[] performExecute(PortObject[] inObjects, ExecutionContext e
...
+      if(metadata != null && SwaggerUtil.getLanguageWrittenIn(metadata) != null) {
+        String languageWrittenIn = SwaggerUtil.getLanguageWrittenIn(metadata);
+        String pythonPath = null; // Path to the Python executable in the conda environment
+        String envPath = null; // Path to the conda environment
+        if(PreferencePage.getPython3CondaEnvironmentDirectoryPath() != null)
+          envPath = PreferenceInitializer.getPython3Env();
+        else
+          envPath = PreferenceInitializer.getPython2Env();
+        
+        if(((FskPortObject)inObjects[0]).overridePackages) {
```
schuelet, line 543: Throws an error if the Editor has no input
image

`ERROR FSK Editor JS        3:1        Execute failed: Cannot read field "overridePackages" because "inObjects[0]" is null
`

schuelet (changes requested):
@ahmadswaid Maybe we need to debug some more for Windows OS, also in fskml (PythonScriptAnalyser): 
image

There, in line 148, the command for the ProcessBuilder contains an item from scriptArguments that is not a valid process (it is the python env path):
`[C:\Users\thsch\anaconda3\envs\py39_fsk_env\python.exe, C:\Users\thsch\analyze_script.py, C:\Users\thsch\AppData\Local\Temp\239328783802838195.py, C:\Users\thsch\anaconda3\envs\py39_fsk_env]`
The last item will probably throw the error because it is just a directory.

de.bund.bfr.knime.fsklab.nodes/src/de/bund/bfr/knime/fsklab/v2_0/editor/FSKEditorJSNodeModel.java
schuelet: Another thing: the python scripts are stored in the user folder. We should either put them in the workflow folder or use the KNIME python-integration way to send commands to the python server instead of directly calling a python process.

